PR #4, opened 2014-04-15
Add Glfw.Windows.Window.Set_Cursor_Hidden 

fawick:
Adds the ability to hide the mouse cursor for a window

I am not very familiar with Ada so I took the liberty of not wrapping the original GLFW3 API but wrote a dedicated procedure for hiding the cursor instead.


flyx:
Fixed by #8 .

PR #5, opened 2015-05-24
add missing libraries to linker flags

Beowulf0:


onox:
Why is this needed? Without the patch, the test binaries still link to the libraries automatically.

flyx:
It may differ between Linux distributions. Since the linker flags do no harm, I included them because Beowulf0 apparently could not compile without them. The linked libraries are dependencies of GLFW; I am not sure why they sometimes aren't needed.

Sadly, the `.gpr` files are not that portable. Names for libraries may differ between distributions, and dependencies may change. You might be better off writing a Makefile and using `pkg-config --libs glfw3` anyway. The `.gpr` files are mainly for an easy start.

PR #28, opened 2017-01-10
New examples

rogermc2:


rogermc2:
I'm a bit baffled.
Git seems to put the code for this request into a new directory ?

flyx:
Well one directory is named `example` while the other is named `Example`. Mind the casing.

rogermc2:
Yes. That's what I noticed.
But I don't know where the E came from.
My directory structure here is
OpenGLAda
   examples
      Moving_Triangle
      Simplest

Its as if GitHub Desktop changed examples to Examples.
So, at the moment, I can't figure out what to do.
Looking at History in GitHub Desktop, the directories are definitely shown as
examples/Superbible/Moving_Triangle/src/

So, where did the capital E come from?

rogermc2:
I deleted this and opened a new one, #29 then committed and pushed my Moving_Triangles files.
But I still see an Examples and an examples directory.
Should I change the name of my examples directory to Examples?

rogermc2:
OK. Confusion on my part yet again
Somehow, I got the titles mixed up. I've now changed my local directory name to Examples.

PR #31, opened 2017-01-13
More Examples

rogermc2:
Attempt to add a new directory to Examples/Superbible.


rogermc2:
I am stuck again!
I have made a branch More_Examples and added a folder Moving_Triangle to my /Ada_Source/OpenGLAda/Examples/Superbible folder but no matter what I try Moving_Triangle files do not show up in GitHub Desktop as commits. It just states "No uncommitted changes".
On previous attempts, when I delete the branch, the  Moving_Triangle files disappear from the Moving__Triangles directory so they are obviously "known" to the branch.
I seem to be missing something fundamental to using GitHub Desktop!
Please advise advise me.
Thanks

flyx:
Are you on Windows? If so, the problem is that you have added `moving_triangle` to `.gitignore` which tells git that *everything named `moving_triangle` should be ignored*. That includes your folder `Moving_Triangle` because Windows paths are case insensitive. The solution is to remove `moving_triangle` from `.gitignore` and add it to a new `.gitignore` *inside the `Moving_Triangle` folder*. This way, only the resulting binary is ignored.

rogermc2:
THANKS!
So obvious now that you have pointed it out.
I have now corrected that the commited and pushed the Moving_Triangles source files and gpr.

flyx:
I did some restructuring on your code. Check out current master!

 * I moved all directory names to lower case. This will hopefully avoid confusing about different names in the future.
 * I moved gitignores for the example binaries into their folders to avoid confusion with directory names.
 * I moved common code into the `common` folder under `superbible` so that it is not replicated in every example. On my way, I made some necessary enhancements.
 * I simplified some of the code, for example I removed packages with only one procedure; they are now a standalone procedure.
 * Returning a `Boolean` to tell the caller whether an operation succeeded is bad style. I replaced it with an exception (see code in `Program_Loader`).

Some comments for you:

 * Improve error handling. You have exception handlers at multiple places that do the same – they output some debug information *but do not exit the program*. This is a bad thing because it means that succeeding code may be executed, which will probably lead to more errors. A good thing to do would be to move the exception handlers up to the main loop or even the entry point procedure.
 * There is some more code replicated in the `Main_Loop` of the two example. Perhaps that too can be moved to `common`.

rogermc2:
Thanks flyx.
I'll update my other examples that I have ready for submission for consistency with your changes.
With regard to exceptions. I had wondered about whether they should be kept in the submitted code.
I definitely need them in many procedures to help with development but obviously they shouldn't be in "delivery" code.
Currently I have dug myself a big hole trying to convert some SOIL C code to Ada.

flyx:
Well, if you move the exception handlers up to `Main_Loop` or even the entry procedure, you would still get the same information with `Exception_Information`. There is no need to have the exception handlers inside the procedure where it is raised. Another advantage would be that the exception handler code is not duplicated, but all in one place.

If you need additional information about where the exception occurred, you can do this in the procedure:

    exception
       when others =>
          Put_Line ("Exception in Render"); raise;

This way, you get the name of the procedure as output, but the handler will re-raise the exception, which will be handled later on.

It is also possible to get a stacktrace from the exception if the code is compiled properly. If you use that, you do not need the `Put_Line`. See [this Ada Gem](http://www.adacore.com/adaanswers/gems/gem-142-exceptions/).

rogermc2:
Thanks for this Flyx,
	I'll definitely make use of it in future. It certainly looks like it satisfies my needs nicely.
Regards,
Roger

PR #32, opened 2017-01-24
Add Vertex Colouring example

rogermc2:
colour_vertices now ignored by local .gitignore file


.gitignore
```diff
@@ -2,4 +2,10 @@ lib
...
 .DS_Store
-.#*#
\ No newline at end of file
+<<<<<<< HEAD
+.#*#
+a_dot
+moving_triangle
+=======
+.#*#
+>>>>>>> flyx/master
```
flyx: This is a conflict you failed to solve after merging master into your branch. Only lines 6-8 should stay.

examples/Superbible/Vertex Colouring/.gitignore
```diff
@@ -0,0 +1,3 @@
+obj
```
flyx: `obj` is already ignored by the top level `.gitignore` and does not need to be repeated here.

examples/Superbible/Vertex Colouring/src/main_loop.adb
```diff
@@ -0,0 +1,93 @@
...
+    subtype tVec4f is GL.Types.Singles.Vector4;
+
+    Rendering_Program : GL.Objects.Programs.Program;
+    Vertex_Array      : GL.Objects.Vertex_Arrays.Vertex_Array_Object;
+
+    --  ------------------------------------------------------------------------
+
+    procedure Render_Triangle (Current_Time : Glfw.Seconds) is
+        use Math_Functions;
+        Back_Colour : GL.Types.Colors.Color := (0.5 * (1.0 + Sin (Single (Current_Time))),
```
flyx: The code would be more readable if you calculated `Single (Current_Time)` only once and stored it in a variable.

.gitignore
rogermc2: I don't really understand the Git message in this file.
I've hopefully fixed it by copying the contents of the flyx/master file.
What is the .#*# for?

examples/Superbible/Vertex Colouring/src/main_loop.adb
rogermc2: Agreed!
Have fixed.

flyx:
renamed the paths to all-lowercase after merging.

PR #41, opened 2017-02-27
Durian Textures

rogermc2:
Finally, a working image texture example!


rogermc2:
I don't understand why so many files were added in addition to my hello_textures files?

flyx:
You mean why you have that much commits in on pull request? That's because I usually squash your PRs, i.e. make them one commit when merging them. This means that afterwards, there is one commit which does all the changes you did, but your original commits are not getting included. So each time you do another pull request, GitHub correctly shows all your commits which are not directly part of my branch – even though the *changes* those commits made are part of my branch. This is how git works, and it does no harm.

You can get rid of those commits by creating a new branch based on my master branch. Currently, you are always merging my master branch into your branch, which preserves your old commits.

(I will process this PR later when I have more time)

rogermc2:
Thanks for the explanation.
I'll try to remember to branch from your master in future.

PR #42, opened 2017-02-28
OGL Triangle

rogermc2:
New OGL Tutorials directory with first example.
I origonally pulled this into my own master by mistake!


flyx:
The `common` stuff is just copied from the other examples, right? So it would probably make sense to move the `common` folder up right into `examples` to avoid code duplication.

rogermc2:
Agreed.
I'll have a go at doing that tomorrow.
I guess you'll hold of on this one until I've done that?

flyx:
Yes, I'll wait. Please remove all other `common` folders, reference the new one instead and make sure everything compiles.

flyx:
btw, I am currently working on generating a mapping table from OpenGL function names to OpenGLAda subroutines, see [here](http://flyx.github.io/OpenGLAda/mapping.html). This will make it easier to translate C OpenGL code to Ada, but it is currently very incomplete.

rogermc2:
I have also been doing the same.
Definitely helps.
See attached.

Regards,
Roger

rogermc2:
Looks really good.
Easy to find functions using simple Find command.
Regards,
Roger

rogermc2:
I pushed changes for using the examples/common directory but I don't see them  in the list above.

PR #43, opened 2017-03-01
Common

rogermc2:
Replace common directories with a single common directory at the top level of Examples.


rogermc2:
I hope I've covered everything.
Particularly in the other, ogl_tutorials, PR which was a bit of a messy operation.

PR #44, opened 2017-03-02
Update gitignore to ignore hello application

rogermc2:


flyx:
I do not see why I should merge this PR with just that one change.

PR #46, opened 2017-03-02
Matrices

rogermc2:
Add totrial_03


flyx:
Again, this contains multiple examples and duplicated `common` code, see my comment on the other PR.

PR #52, opened 2017-03-05
Matrices example

rogermc2:
I had some trouble resolving the .gitignore conflict which, presumably, should be a simple thing to do.
I'd really appreciate a set of instructions on how to resolve such conflicts.


examples/ogl_tutorials/tutorial03_matrices/src/maths.ads
```diff
@@ -0,0 +1,34 @@
+
+with Ada.Numerics;
+with Ada.Numerics.Generic_Elementary_Functions;
+
+with GL.Types;
+
+package Maths is
```
flyx: I do not quite understand why this package exists. `+`, `-` and `*` are well-defined for both `GL.Types.Singles.Vector3` and `GL.Types.Singles.Matrix4`, as defined in `GL.Vectors` and `GL.Matrices` respectively. This example confuses users as it suggests you need to write your own implementation for these mathematical functions although they already exist. Please remove them or explain why you cannot use those from OpenGLAda.

rogermc2: OK. Understood.
I would only have declared these functions because of some compilation complaint and certainly would have preferred not to declare them if I had realized an alternative solution at the time.
I suspect I initially declared the subtypes as new types but subsequently changed them to subtypes to solve some other problem.
Based on your advice, I'll remove the mathematical functions and solve any problems that arise without recourse to them.
The subtypes are provided for consistency with types in a set of tutorials.
If you prefer I can remove them.
I need to retain the package to declare the two procedures. (Unless there are equivalent OpenGLAda procedures that I haven't discovered.)
I'll also remove the unnecessay "private" which previously escaped my attention.

flyx: > The subtypes are provided for consistency with types in a set of tutorials.
> If you prefer I can remove them.

I would say that if the subtypes only exist to have type names similar to C, they should be removed, since this is not what subtypes in Ada are for. The subroutine names in OpenGLAda also differ and rightfully so (to comply to the *Ada Quality and Style Guide*), so having different names for the vector and matrix types is not a problem.

examples/ogl_tutorials/tutorial03_matrices/src/vertex_data.ads
```diff
@@ -0,0 +1,29 @@
...
+   Default_Terminator : GL.Types.Single := 0.0;
+
+ --  package Interfaces.C.Pointers generic interface:
+ --  type Index is (<>);          : GL.Types.int
+ --  type Element is private;     : GL.Types.Single
+ --  type Element_Array is array (Index range <>) of aliased Element;
+ --  Default_Terminator : Element;
+
+   package pVertex_Pointers is new Interfaces.C.Pointers
+     (GL.Types.Int, GL.Types.Single, tElement_Array, Default_Terminator);
```
flyx: Is there a specific reason for defining this instead of just using `GL.Types.Single_Pointers`?

examples/ogl_tutorials/tutorial03_matrices/src/maths.ads
rogermc2: OK.
Based on your comment, I'll remove the subtypes.

examples/ogl_tutorials/tutorial03_matrices/src/vertex_data.ads
rogermc2: I'll try and use GL.Types.Single_Pointers instead.

rogermc2:
* are well-defined for both GL.Types.Singles.Vector3 and GL.Types.Singles.Matrix4

But :
    MVP_Matrix          : GL.Types.Singles.Matrix4;

        Model_Matrix      : GL.Types.Singles.Matrix4 := Singles.Identity4;
        Projection_Matrix : GL.Types.Singles.Matrix4 := Singles.Identity4;
        View_Matrix       : GL.Types.Singles.Matrix4 := Singles.Identity4;

        MVP_Matrix := Projection_Matrix * View_Matrix * Model_Matrix;
produces:
there is no applicable operator "*" for type "Matrix4" defined at gl-algebra.ads:78, instance at gl-types.ads:114

Am I doing something wrong?
Do you need a minimal test program?

flyx:
Did you put in `use type GL.Types.Singles.Matrix4`?

rogermc2:
Quite weird!
I hadn't used use type GL.Types.Singles.Matrix4
I tried it in a number of places; finally at the start of Maths.ads just before the package declaration which solved the problem.
HOWEVER, I then deleted it and the program still compiled and worked.
I then cleaned the project, rebuilt and it still worked without use type GL.Types.Singles.Matrix4.

Although it currently compiles OK, I presume from your comment that I should add
use type GL.Types.Singles.Matrix4?

Aside from this, could you please explain why use type GL.Types.Singles.Matrix4 should be needed?
As normally, I expect that    Model_Matrix : GL.Types.Singles.Matrix4;   would be sufficient.

rogermc2:
I made your recommended changes to Maths.
However now, when I multiply three identity matrices together, I don't get an identity matrix back.

        use GL.Types;
        use Maths;
        use type GL.Types.Singles.Matrix4;

        Result                   : GL.Types.Singles.Matrix4;
        Model_Matrix       : GL.Types.Singles.Matrix4 := Singles.Identity4;
        Projection_Matrix : GL.Types.Singles.Matrix4 := Singles.Identity4;
        View_Matrix          : GL.Types.Singles.Matrix4 := Singles.Identity4;

       Result := Projection_Matrix * View_Matrix * Model_Matrix;
       for row in GL.X .. GL.W loop
            for col in GL.X .. GL.W loop
                Put (Single'Image (Result (row, col)));
            end loop;
            New_Line;
       end loop;

produces Result:
 1.00000E+00 2.00000E+00 3.00000E+00 4.00000E+00
 5.00000E+00 7.00000E+00 9.00000E+00 1.10000E+01
 1.30000E+01 1.50000E+01 1.80000E+01 2.10000E+01
 2.40000E+01 2.70000E+01 3.00000E+01 3.40000E+01

rogermc2:
I've pushed the current changes.
These include the identity multiplication test as I need to get this right before returning to the final code and wanted to get all of the major changes incorporated.

examples/ogl_tutorials/tutorial03_matrices/src/vertex_data.ads
rogermc2: Changed to GL.Types.Single_Pointers.

flyx:
> I can't seem to get vector multiplication (V1 * V2) to work so have retained my dot and cross function routines for vectors.
> Are there OpenGLAda functions for these? I can't seem to find them.

No, but there should. Can you move your functions into `GL.Vectors`, renaming them to `Dot_Product` and `Cross_Product`?

> However now, when I multiply three identity matrices together, I don't get an identity matrix back.

That's a bug, I will investigate.

flyx:
I pushed a fix to the matrix multiplication; merge my master branch and try again.

rogermc2:
Great!
The matrix multiplication fixed worked and the program displayed the expected image.

rogermc2:
I'll move the functions into GL.Vectors, renaming them to Dot_Product and Cross_Product.
However, first, I wondered if the Cross_Product might be better implemented as "*" for consistency with
a) Your matrix "*" function and
b) Your other vector "*" functions?

flyx:
`*` was always a sorry excuse for typesetting a mathematical operator, because a better alternative was not available in ASCII. Thus, I hesitate to use it in any situation where its meaning is not obvious.

The typical mathematical notation for the cross product is „×“, while the typical notation for the dot product is „⋅“. Since it is not obvious which notation the asterisk should map to, I do not want to use it for either. In the other vector functions, it is obvious, since there is only one *product* operation defined for a vector and a scalar.

For matrices, on the other hand, the term *matrix product* is well-defined. While there also exist various other product operations like the Kronecker product you mentioned or the Hadamard product which I referred to as *element-wise* in my comment, they both have a clear distance in naming and notation from the matrix product. For example, you frequently drop the operator for matrix products in mathematical notation, which I see seldom done for dot or cross products of scalars. For these reasons, I deem it reasonable to have a `*` operator on matrices, but go for `Dot_Product` and `Cross_Product` for vectors.

rogermc2:
Fine.
Dot_Product and Cross_Product for vectors it is.

flyx:
Just put it into `examples/common`. It doesn't really matter that the other example sets do not use it.

rogermc2:
I've had a go at moving the vector product declarations into gl-vectors.ads but don't have experience implementing generics.
My first attempt in which I don't know how to declare Dot_Product's Single return type:
   function Cross_Product (Left : Vector; Right : Vector) return Vector;
   function Dot_Product (Left : Vector; Right : Vector) return Single;

It would seem to require something like 
type Return_Type is private;
   function Dot_Product (Left : Vector; Right : Vector) return Return_Type;
Please advise the correct way ahead.
Thanks

rogermc2:
I've implemented the dot product and cross product functions in GL.Vectors and pushed the changes.
I only recalled after starting on the move that the cross product is only defined for 3D.
I've checked that the move works for the Matrices program.
I look forward to any improvements you might recommend as this is my first experience in generic design.

rogermc2:
Testing revealed problems with cross_product which I have now corrected.
The Matrices example now appears to be working correctly.
I inadvertantly made a commit entitled "se enter the commit message for your changes. Lines starting" which I thought I had removed before pushing.

flyx:
While there are generalizations available for the cross product in any ℝ^n where n ≥ 2, I think that it does not really make sense to implement them. I did not think about the cross product only applying to three-dimension, which makes it unwise to add it to `GL.Vectors`. Your code now throws a runtime exception for non-3-dimensional vectors instead of disallowing the operation altogether. Please move the `Cross_Product` to `GL.Algebra` where it can be defined only for `Vector3`.

Your `Dot_Product` implementation looks good.

rogermc2:
While there are generalizations available for the cross product in any ℝ^n where n ≥ 2?
Do you have a reference for n>3.  I'm quite interested for comparison with geometric algebra in which the GA outer (wedge) product is defined for all dimensions.
I'm supposing that the determinant format for the 3D outer product might possibly be used to define higher dimensional cross products. Out of interest, I'll try and see if such an approach makes sense (on paper).
To provide some confidence that Cross_Product and Dot_Product are implemented correctly, I checked them out by reproducing the algorithm for one of the Math.adb procedures (Init_Lookat_Transform) in Mathematica. The results agreed with each other.

OK, I'll have a go at moving Cross_Product to GL.Algebra defining it for Vector3 only.

rogermc2:
After some thought I realized that  there should be generalizations  of the 3D cross product to higher dimensions; a cross product vector in n dimensions being the vector orthogonal (normal) to the hyperplane surface formed by n-1 non-parallel vectors.

rogermc2:
I have moved Cross_Product to GL.Algebra where it is defined only for Vector3.
The definition of Vector3 in GL.Algebra only allows homogenous coordinates (X, Y, Z), therefore this Cross_Product is restricted to 3D vectors specified with homogenous coordinates.
Is there a way of extending to generalized coordinates?

flyx:
IIRC *(X, Y, Z)* are not homogeneous coordinates. *(X, Y, Z, W)* are – which are used for `Vector4` (hence the type `Index_Homogeneous`). It would make sense to make the `Cross_Product` also available for `Vector4`, but I am not deep enough in this kind of math to know how that is defined. If you know, feel free to implement it.

rogermc2:
I agree that (X, Y, Z) are not homogeneous coordinates.
I should have stated something like "a subset  (X, Y, Z)  of  homogeneous coordinates".
I'll look into the possibility of making the Cross_Product also available for Vector4.
I have an idea of what to do but I'll try and track down correct mathematical theory before attempting any coding.

rogermc2:
I've done some research into the cross product.
It seems that under a generally accepted mathematical definition cross products are only possible for dimensions 3 and 7.
I think that the strict mathematical requirements involve cross products suitable for geometry.
Relaxation of the requirements does seem to allow for suitable definitions of cross product where the strictness is not required.
Because of the strictness condition I was going to suggest not implementing a 4D cross product.
However, Mathematica does provide a cross product for dimensions other than 3.
This "Cross in dimension n is the contraction of n-1 vectors into the Levi-Civita tensor".
Basically, Mathematica's Cross returns a 4D vector that is the cross product of three 4D vectors.
This Mathematica implementation turns out to be the negative of a cross product calculated by the determinant method.
From what I've read, I think that it needs to be kept in mind that any 4D cross product should not be interpreted from a geometrical point of view.
It should be fairly straight-forward to implement Mathematica's 4D cross product.
So, just let me know if you feel that its appropriate. If so, I'll implement it.

flyx:
Well, OpenGLAda is not and does not want to be a mathematical library. The vector and matrix operations are mostly provided because it is very likely that they will be needed. The question we need to answer is thus, is the cross product for `Vector4` needed in the context of OpenGL? I think it may be, because a `Vector4` is basically a `Vector3` transformed into a homogeneous coordinate system, and that does not (or should not) make the operation invalid.

However, it may well be that mathematically, there is no matching operation for the cross product in 4 dimensional space. To be sure to get a cross product that matches the semantics of the 3d one, we could implement it as

 * normalize the 4d vector (so that `W = 1`).
 * calculate the normal 3d cross product on `(X, Y, Z)`
 * add `W = 1` to the result and return the new 4d vector.

This will of course fail if `W = 0`, but afair this should never happen when dealing with homogeneous coordinates.

I have no strong opinion on whether to implement this or not. If you feel like you want to add it, do it (but add a documentation comment about what it does). If not, leave it.

rogermc2:
Your suggested implementation makes mathematical sense to me in my limited knowledge of OpenGL (or computer graphics in general).
In one sense, because the cross product is restricted to what I regard as the geometrical elements of the 4D vector; I regard the W component as "informational". Again, this is based on my limited understanding and not wanting to spend a lot of time on a deeper investigation.
Incorporation in this context seems advantageous from a completeness point of view, particularly if it would be useful.
On the other hand I would only want to incorporate it if you're sure your suggested implementation would produce "correct" results or if you can provide a suggestion as to some tests to prove "correctness".
I think that the W = 0 situation would need to be catered for but I need to know your preferred method  (I gather that you may not be keen on exceptions in these procedures).
Again, in the context of your suggested implementation, completeness seems advantageous to me.
If I did add it, would a documentation comment be required other than as a documentation header in the code?

flyx:
Hmm. What's the matter with the `Perspective` function in `maths.adb`? It is never used.

rogermc2:
The Perspective function in maths.adb is a function that I needed for some earlier work and may be needed for examples that I've yet to provide.
I'd prefer to leave it in for the time being; but will delete it if you prefer.

flyx:
I'd prefer you adding it to `maths.ads` so that it is callable from other packages ;).

rogermc2:
OK.
Also, To_Radians, To_Degrees, Length and Normalize?

flyx:
`To_Radians` and `To_Degrees` are fine for package-intern stuff, but if you plan to use them publicly, you should make them take / return new types `Radian` and `Degree` instead of `GL.Types.Single`. That will ensure that the user always knows which kind of angle to provide. Also, names should be simply `Radians` and `Degrees` conforming to the [Ada Style Guide](https://en.wikibooks.org/wiki/Ada_Style_Guide/Readability#Names):

> Use nouns for non-Boolean functions.

I really like that style guide, so I try to adhere to it.

I thought about incorporating `Length` and `Normalize` directly into `GL.Vectors`, but I opted against it because they only make sense for floating point types. You can make them public if you want.

rogermc2:
OK.
I'll also add some other generally useful OpenGL maths functions, most of which I expect to use in forthcoming examples.
I agree that adhering to the style guide is very advantageous; so, your comments concerning it are very welcome.

PR #54, opened 2017-03-14
Square example

rogermc2:


flyx:
The directories `open_gl` and `durian` in `examples` are missing information about where these examples are from (like the Readme files in `ogl_tutorials` and `superbible`). Those would be nice to add.

rogermc2:
OK. I intend to add these shortly, probably in an update to Maths.

PR #55, opened 2017-03-17
Maths update

rogermc2:


examples/common/maths.ads
```diff
@@ -1,11 +1,32 @@
-with GL.Types;
+
+with GL.Types; use GL.Types;
 
 package Maths is
+
+    subtype tDegree is Single;
+    subtype tRadian is Single;
```
flyx: Subtypes are implicitly convertible into each other. To actually improve type safety, these both should be `new Single` types. Also, I don't like the `t` prefix. I have seen it used in Pascal and Visual Basic, but never in Ada. Simply `Degree` and `Radian` suffices.

If you named the types like that because of the conversion functions, they can be named either `Degrees` / `Radians` (plural; ) or perhaps `As_Degree` / `As_Radian`.

examples/common/maths.adb
```diff
@@ -109,49 +103,104 @@ package body Maths is
...
-
-    function Perspective (Top, Bottom, Left, Right, Near, Far : Single)
+    --  Perspective_Matrix is derived from Computer Graphics Using OpenGL
+    --  Chapter 7, equation 7.13
+    function Perspective_Matrix (Top, Bottom, Left, Right, Near, Far : Single)
                           return GL.Types.Singles.Matrix4 is
         use GL;
+        dX : Single := Right - Left;
+        dY : Single := Top - Bottom;
+        dZ : Single := Far - Near;
```
flyx: These can be `constant`.

examples/common/maths.ads
```diff
@@ -1,11 +1,32 @@
...
+    procedure Init_Orthographic_Transform (Bottom, Top, Left, Right,
+                                           Z_Near, Z_Far : Single;
+                                           Transform     : out Singles.Matrix4);
     procedure Init_Perspective_Transform
-      (View_Angle, Width, Height, Z_Near, Z_Far : GL.Types.Single;
-       Transform                                : out GL.Types.Singles.Matrix4);
+      (View_Angle, Width, Height, Z_Near, Z_Far : Single;
+       Transform                                : out Singles.Matrix4);
+    function Length (V : Singles.Vector3) return Single;
+    procedure Normalize (V : in out Singles.Vector3);
```
flyx: Since normalization is a mathematical function, I would prefer this to be a function named `Normalized` which returns a new `Vector3`. That is more versatile as it can be used for variable initialization and such.

rogermc2:
OK. Hopefully I've made all the modifications. Hoping to push shortly for further inspection.
"integrating some of these functions into OpenGLAda itself " is a great idea.
But perhaps wait until I've given them a good workout.
Personally, I think a good unit testing phase is needed before wider use.

flyx:
While unit testing in general is a good idea, I prefer a good code review for mathematical functions. Unit testing methology is more fit for checking state changes (i.e. *is this thing in the DB after I execute that function*); mathematical calculations are best checked by reading and comparing them to the mathematical definition.

rogermc2:
"mathematical calculations are best checked by reading and comparing them to the mathematical definition."
Agreed.
I would add "but not by the author". That is, a good code review by an independent auditor.
However, I am doing some checking by implementation in Mathematica for the uses that I put these routines to. This is prompted by difficulties I am having with the example that I am currently trying to implement.
I think this refers to your i.e. That is, do the uses I put these functions to produce the expected result?

rogermc2:
I'm not confident of the Init_Lookat_Transform, so best hold off implementing for a while.
There seem to be two versions in use, the basic difference being that one 3x3 scaling/rotation sub-matrix is the transpose of the other.
I don't think this should be a problem as presumably, the difference is simply rotation direction, CW or CCW.
After a fair bit of Googling, I still not sure if one is preferred over the other.
OpenGL SuperBible prescribes the opposite to my currently implemented version provided by Computer Graphics Using OpenGL which is a much older book than OpenGL SuperBible.
My current inclination is to switch to the OpenGL SuperBible version as it is the more recent book.

rogermc2:
I am checking the program results against Mathematica implementations which gives some confidence.
Also there seem to be a couple of differences between the  Computer Graphics Using OpenGL and OpenGL SuperBible.

rogermc2:
Having essentially completed my derivation, on comparison with my previous efforts I have discovered that I have been misreading the textbooks.
I had been interpreting s.x  to mean s(x) instead of s dot x!
Hopefully repairs won't take too long.
My derivation, based on a mathematics article on the subject corresponds to the OpenGL SuperBible so I'll use that and change other source references to OpenGL SuperBible references where possible.

rogermc2:
I am now fairly confident that the maths algorithms are correct.
All routines are based on Computer Graphics Using OpenGL equations.
The main concern was the lookat matrix which I have taken from Computer Graphics Using OpenGL with a correction of (W,W) = 1 instead of 0 (I presume a typographical error).
I believe the OpenGL SuperBible lookat matrix to be incorrect, particularly in the translation column (W) which uses model coordinates instead of projecting the model's position vector onto the camera axes.
Another difference is that the rotation (3x3) submatrix in SuperBible is the transpose of the one in Computer Graphics Using OpenGL. I assume that this would only change the direction of rotation and probably mean the difference between left and right handed coordinate systems. My understanding is that the model system is right handed while the camera system is left handed.
I'm not that confident that I can distinguish between the two but, so far, my feeling is that the Computer Graphics Using OpenGL version is correct.

rogermc2:
Hold off!
Suspect perspective algorithm problem.

rogermc2:
Well, I've checked against a number of sources and the perspective algorithm seems OK.

rogermc2:
This is a great help. Thanks.
I think one of my comments against SuperBible 's translation components was incorrect so have amended that comment accordingly.

PR #60, opened 2017-04-04
Math utiliy enhancement.

rogermc2:
Add print 3D vector and print OpenGL 1D array procedures.
Make Maths.Single_Math_Functions package visible as it is needed for many examples.


examples/common/maths.adb
```diff
@@ -206,6 +204,45 @@ package body Maths is
     end Scaling_Matrix;
 
     --  ------------------------------------------------------------------------
+
+    function To_3 (Array_4 : Single_Array) return Single_Array is
```
flyx: The name of this function does not tell at all what it does. Also, I do not really understand the purpose of this function. It seems like you actually want to transform a `GL.Types.Singles.Vector4_Array` to a `GL.Types.Singles.Vector3_Array`; but instead of normalizing to `W = 1`, you just drop the fourth component. But this is just guessing the intention; it seems very odd to use `Single_Array` instead of the proper vector array type for this.

examples/common/maths.adb
```diff
@@ -206,6 +204,45 @@ package body Maths is
...
+        Array_3  : Single_Array (0 .. Array_4'Length * 3 / 4 - 1);
```
flyx: Ada arrays start at `1` by convention; your code would be simpler if you adhere to that (you can drop the `- 1`).

examples/common/maths.adb
```diff
@@ -206,6 +204,45 @@ package body Maths is
...
+        Index_3  : Int;
+        Index_4  : Int;
+    begin
+        Index_3 := 0;
+        Index_4 := 0;
```
flyx: This assumes `Array_4` starts at `0` which it is in no way required to do.

examples/common/maths.adb
```diff
@@ -206,6 +204,45 @@ package body Maths is
...
+        while Index_4 < Array_4'Length - 1 loop
```
flyx: Even if you assume `Array_4` starts at `0`, this condition is wrong since you access `Array_4 (Index_4 + 2)` inside the loop body.

examples/common/maths.adb
```diff
@@ -206,6 +204,45 @@ package body Maths is
...
+            Array_3 (Index_3 + 2) := Array_4 (Index_4 + 2);
+            Index_3 := Index_3 + 3;
+            Index_4 := Index_4 + 4;
+        end loop;
+        return Array_3;
+    end To_3;
+
+    --  ------------------------------------------------------------------------
+
+    function To_4 (Array_3 : Single_Array) return Single_Array is
```
flyx: This function has all the same problems as `To_3`.

examples/common/utilities.adb
```diff
@@ -11,6 +11,13 @@ with GL.Types.Colors;
 
 package body Utilities is
 
+   procedure Clear_All (Colour : GL.Types.Colors.Color) is
+   begin
+        GL.Buffers.Clear ((True, True, True, True));
+        GL.Buffers.Set_Color_Clear_Value (Colour);
```
flyx: Shouldn't these two calls be the other way round? As it is, you first clear the buffers with the previously set clear color, and then you set the new one.

examples/common/utilities.adb
```diff
@@ -27,6 +34,46 @@ package body Utilities is
 
     --  ------------------------------------------------------------------------
 
+    procedure Print_GL_Array3 (Name : String; anArray : GL.Types.Single_Array) is
```
flyx: I still don't get why you are not simply using `GL.Types.Singles.Vector3_Array`.

examples/common/utilities.adb
```diff
@@ -27,6 +34,46 @@ package body Utilities is
...
+            Vec_3 (Z) := anArray (Index + 2);
+            Print_Vector (Int'Image (Index) & ":  ", Vec_3);
+            Index := Index + 3;
+        end loop;
+        New_Line;
+    end Print_GL_Array3;
+
+    --  ------------------------------------------------------------------------
+
+    procedure Print_GL_Array4 (Name : String; anArray : GL.Types.Single_Array) is
```
flyx: ... or `GL.Types.Singles.Vector4_Array`.

examples/common/maths.adb
rogermc2: OK.
Thanks for your comments.
These seemed to do I wanted but, in light of your comments, I'll remove them and use your comments to hopefully do a better job.

examples/common/maths.adb
rogermc2: My understanding is that OpenGL arrays, particularly vertex arrays start at 0.
I agree that Ada  arrays start at 1 by convention and this is certainly my preference but I assume that GLSL shaders expect arrays to start at 0.
Any way, as I said above I'll remove these functions from Maths until and if I can fix them .
By the way, I have just updated my OS to Sierra.

examples/common/utilities.adb
rogermc2: OK, I'll check this out.
I may have misunderstood what GL.Types.Singles.Vector3_Array is.
Certainly would be better if I can use it.

examples/common/utilities.adb
rogermc2: I still really don't understand how these work.
I did these in the same order as the Clear Buffer Flag and Colour routines which I thought I had ordered in agreement with a previous recommendation of yours.
As I understand, Clear clears four bits of a "bitfield"; it implements glClear.
Set_Color_Clear_Value implements glClearColor.
I'll try and do some more research to understand this better. That is, see if the OpenGL documentation on glClear and glClearColor clarifies my understanding.

rogermc2: Yes.
I checked the OpenGL documentation and now understand the process.
Set_Color_Clear_Value sets the value to which a buffer should be set when cleared.
Clear implements glclear which "clears" selected values to the previously selected value set by Set_Color_Clear_Value .

examples/common/utilities.adb
rogermc2: I'd like to use GL.Objects.Buffers.Load_To_Buffer with GL.Types.Singles.Vector3_Array.
If this is possible, how to do it?
So far, I haven't been able to figure it out.

examples/common/utilities.adb
rogermc2: Fixed.

flyx:
Frankly, I don't really understand why I require an instantiation of `Interfaces.C.Pointers` for `Load_To_Buffer`. Its implementation only uses the `Element` and `Element_Array` types from it, so I could just require those two types as generic parameters.

I wonder if it's too late to change that.

rogermc2:
I'm not quite sure what you mean.
I wonder if it's too late to change that? Never too late from my point of view.
If you need to change the implementation, its not too late from my point of view.
Is there something that I need to do?

flyx:
Well, what I'd like to do is to change

```Ada
generic
   with package Pointers is new Interfaces.C.Pointers (<>);
procedure Load_To_Buffer (Target : Buffer_Target;
                          Data   : Pointers.Element_Array;
                          Usage  : Buffer_Usage);
```

to:

```ada
generic
   type Index is (<>);
   type Element is private;
   type Element_Array is array (Index range <>) of aliased Element;
procedure Load_To_Buffer (Target : Buffer_Target;
                          Data   : Pointers.Element_Array;
                          Usage  : Buffer_Usage);
```

This would break all existing code that uses it, including a lot of your examples, which would need to be fixed.

rogermc2:
That's OK.
I'm happy to fix things for the sake of overall improvement.

rogermc2:
I have completed a few more examples and also implemented a quaternion rotation system.
My quaternion branch includes all updates to maths and utilities plus examples modified to use quaternions.
I'm wondering if we should delete this PR and do all the changes in a PR from my quaternions branch?

rogermc2:
I'm replacing this PR by a quaternian implementation PR which incorporates these and other changes.

PR #63, opened 2017-04-20
Add new example - Coloured cube

rogermc2:
I had some problems getting this branch up to date.
Hopefully its OK now but I'm not very confident about it.


flyx:
The naming inside the `ogl_tutorials` directory is a bit verbose. Would you mind renaming the directories from `tutorial[Number]_[Name]` to just `[Number]_[Name], like `04_coloured_cube`? Also, the gpr files for the first tutorials were named simply `tutorial_[Number].gpr`, this one now is `tutorial04_coloured_cube.gpr`, that seems unnecessary.

rogermc2:
OK. Done.

PR #64, opened 2017-04-21
New textured cube example

rogermc2:


examples/common/io.ads
```diff
@@ -0,0 +1,22 @@
+
+with Ada.Sequential_IO; Use Ada;
+with Ada.Strings.Unbounded; use  Ada.Strings.Unbounded;
+
+with GL.Types; use GL.Types;
+
+package IO is
```
flyx: What is the purpose of this package? The `Read_String` procedures are never used. I also think that using an `Unbounded_String` as `out` parameter is unnecessary; you could simply make it a function and return a normal String.

Why are you using `Byte_IO` instead of `Text_IO`? What's the difference between `Read_String` and `Read_String0` (they seem to do the same)?

examples/common/io.adb
```diff
@@ -0,0 +1,108 @@
...
+    function To_GL_UByte is new
+      Ada.Unchecked_Conversion (Character, UByte);
+    function To_GL_Integer is new
+      Ada.Unchecked_Conversion (Source => tCharacter4_Array,  Target => Int);
+    function To_GL_Float is new
+      Ada.Unchecked_Conversion (tCharacter4_Array, Single);
+    function To_GL_Short is new
+      Ada.Unchecked_Conversion (tCharacter2_Array, Ushort);
+    function To_GL_Unsigned_Integer is new
+      Ada.Unchecked_Conversion (Source => tCharacter4_Array,  Target => uint);
```
flyx: What are these conversion functions for? You do not use most of them, and the one time you do, instantiating `Sequential_IO` with `UByte` would have been a better alternative.

examples/common/load_dds.adb
```diff
@@ -0,0 +1,277 @@
...
+with Ada.Unchecked_Conversion;
+
+with GL.Types;
+with GL.Low_Level.Enums;
+with GL.Objects.Buffers;
+with GL.Objects.Textures.Targets;
+with GL.Objects.Textures.With_2D_Loader;
+with GL.Pixels;
+
+with IO;
```
flyx: This package does not seem to be used.

examples/common/load_dds.adb
```diff
@@ -0,0 +1,277 @@
...
+
+--  Extracted from SOIL.adb  Direct_Load_DDS
+procedure Load_DDS (File_Name  : String;
+                        theTexture : out GL.Objects.Textures.Texture) is
```
flyx: Why was translating this to Ada necessary? Not that I do not want it, but I want to know the reason. I need to review this very carefully because of the usage of `Unchecked_Conversion` (which I may do later).

examples/common/load_dds.adb
```diff
@@ -0,0 +1,277 @@
...
+
+    type tDDS_Header is record
+        Magic          : String (1 .. 4);
+        Height         : GL.Types.UInt;
+        Width          : GL.Types.UInt;
+        Linear_Size    : GL.Types.UInt;
+        Mip_Map_Count  : GL.Types.UInt;
+        Four_CC        : String (1 .. 4);
+        Byte_Size      : GL.Types.UInt := 128; --  bytes
+    end record;
```
flyx: Why is this record named `tDDS_Header` but is loaded via `Load_DSS_Header`? DDS or DSS?

examples/common/load_dds.adb
```diff
@@ -0,0 +1,277 @@
...
+        Dump            : GL.Types.UByte;
+        Texture_ID      : GL.Types.UInt := 0;
+        Header          : tDDS_Header;
+begin
+    Byte_IO.Open (File_ID, Byte_IO.In_File, File_Name);
+    --  Determine file length
+    while not Byte_IO.End_Of_File (File_ID) loop
+        Buffer_Length := Buffer_Length + 1;
+        Read_UByte (File_ID, Dump);
+    end loop;
```
flyx: I hope you are aware that you can just query the size of the file with `Ada.Directories` instead of iterating its whole contents?

examples/common/load_dds.adb
```diff
@@ -0,0 +1,277 @@
...
+
+    Byte_IO.Reset (File_ID);
+    declare
+        Input_Buffer : tUByte_Array (1 .. Buffer_Length);
```
flyx: Reading the content into a buffer and then copying it from there with the `Load_DSS_*` functions is horribly inefficient. Use `Ada.Stream_IO` together with the `'Read` attribute instead to directly read the header and the content into appropriate variables.

flyx:
Also, please remove the binary file.

examples/common/io.ads
rogermc2: Read_String procedures deleted.

examples/common/io.adb
rogermc2: Unnecessary ones deleted.
I'll investigate  instantiating Sequential_IO with UByte.

examples/common/io.ads
rogermc2: Read_String procedures deleted.
I tried converting to Text_IO but was unsuccessful.
However, in trying to do so I somehow made changes that produce a better rendered result  (possibly due  to Adaerizing some indices).
I may try again after attending to other problems.

examples/common/io.adb
rogermc2: Unused ones deleted.
I'll try instantiating Sequential_IO with UByte instead of the other.

examples/common/load_dds.adb
rogermc2: Unused deleted.

examples/common/load_dds.adb
rogermc2: I've gotten rid of all Unchecked_Conversions except for Byte_To_Character as I can't figure out how to otherwise convert a byte to a character which (I thought) should be simple.
As for translating to Ada; two reasons:
1. www.opengl-tutorial.org/beginners-tutorials/tutorial-5-a-textured-cube provides its own loadDDS procedure.
2. My experience with SOIL has been very frustrating. When it fails it gives no real indication of why it fails, only: "raised SOIL.SOIL_ERROR : Unable to open file".
This one was for a png file which I checked exists using  Ada.Directories.Exists.
What is more annoying is that SOIL has been able to read a png file in another example.
I've never had any success with getting SOIL to read DDS files.
I've even tried using the GNAT GPS debugger but have been unable to track the code to the problem area.
I've tried to figure out what might be wrong from the SOIL source but have had no success with that either.
Because of these problems I had started to convert all of SOIL to Ada but the task began to look overwhelming from my point of view requiring more effort than I am prepared to allocate.
In contrast, coding the DDS loader, based on SOIL's DDS loader was not too difficult particularly with the help of the DDS specification and Hex-Editor to inspect an actual DDS file.
However, I don't think I'm prepared to do the same for png files.
Any advice on tracking down the png loading problem will be much appreciated.

examples/common/load_dds.adb
rogermc2: Should be DD. Fixed

examples/common/load_dds.adb
rogermc2: Changed to use Ada.Directories.

rogermc2: Changed to us Ada.Directories.Size.

examples/common/load_dds.adb
rogermc2: Tomorow's task.

rogermc2:
Closing  until current problem fixed.

PR #87, opened 2017-08-31
Ftgl shader example

rogermc2:
This is my first effort at using freetype.
It only displays one character; my simplest approach to getting started with a freetype implementation.
I couldn't figure out what prevents automatic merging.
I look forward to suggestions for improvement.


flyx:
That'll take some time to review ;). Does this code still use parts of FTGL or does it solely depend on FreeType?

rogermc2:
It depends only on FreeType.
I expected that reviewing it will take substantial effort which I will appreciate hope that it will be worthwhile.
In the mean time, I'm starting on the next stage of producing a next level example.

flyx:
I would like to have the FT stuff in a separate project, not in the examples. Can you:

 * Create a project file `opengl-ft.gpr` similar to `opengl-ftgl.gpr` and the others.
 * Create a package `FT` as the root package and make the other packages children of that (e.g. `FT.Glyphs` instead of `FT_Glyphs`). The root package may be a good place for types like `FT_Memory` and `FT_Stream` (also drop the `FT_` prefix from those, they are in a package and thus do not need a namespace prefix).
 * Move the freetype wrapper from the current directory into `src/freetype`.
 * Move all imports into `FT.API` and drop the `FT_` from their names (just like the imported OpenGL functions do not have the `gl` prefix).

Some suggestions from a quick review:

 * Drop the `Get_` prefix from functions like `Get_Bitmap` (in the public packages, not the API imports). Non-boolean functions should use nouns as names according to the [Ada Style Guide](https://en.wikibooks.org/wiki/Ada_Style_Guide/Readability#Program_Unit_Names).
 * Make types private if possible (does the user really need to know `FT_Char_Map` is a `System.Address`?). Drop the `FT_` prefix (see above).
 * `Unsigned_Char_To_Address` is an implementation detail and should be moved to the package body.
 * `Print_Bitmap` should be `Print_Bitmap_Metadata` since it does not actually print the bitmap. Same for `Print_Character_Data`.
 * Add `pragma Pure;` to packages if possible, else `pragma Preelaborate;` if possible.

rogermc2:
Thanks for these which I'll implement as soon as possible.

rogermc2:
I've done my best to implement your recommendations.
Awaiting further comment.

rogermc2:
I've incorporated the recommended changes as best I can.
I've now started on a string example and I am finding the need for many changes so its probably best to postpone further review until I've got that running.
In view of this I'm adding the string example to this branch to ease development.
Should I close this PR until  I've got the string example running?

rogermc2:
Development of my freetype Ada implementation has stabilized so its review can be continued.
've now moved the string example to another branch to ease review of this PR.

flyx:
What is the status of this PR? I cannot review it as long as you keep pushing commits. Are you still working on stuff? If so, please comment here when it's ready (and afterwards, do not push new commits).

rogermc2:
OK.
No more pushes unless changes requested.
Please review.

rogermc2:
I just updated to OSX 10.13 High Sierra following which this and other examples involving freetype failed to work.
The latest commits reflect my attempts to fix the problem.
In doing so I discovered an apparent error in GL.API.spec hence a commit for that.
I now suspect the problem is with the High Sierra upgrades freetype framework as the failing routine appears to be FT_New_Face. I'm now investigating this.

rogermc2:
I have confirmed that the problem is with High Sierra by restoring my second (laptop) computer to Sierra, checking that the restored version of ftgl_shader_example works then pulling the latest remote version and checking that it still works.

flyx:
have you tried linking to a homebrew-supplied freetype instead of the one from the system?

rogermc2:
I tried homebrew freelib 2.8 then did an upgrade which produced 2.8.1 for high_sierra.
The program still failed with both versions of freelib producing the same error.
Under Sierra, all my branches run OK using the latest pulls from origin.

flyx:
I needed to resolve some conflicts after regenerating the API because of changes I did for Windows compatibility. Will probably need to generate again with your additions before merging, please do *not* commit anything to this branch before I'm done.

flyx:
I pushed some improvements to make the interface cleaner. I hope that does not break too much on your side. There are some things I want you to do / answer before merging:

 * Whenever you have an `exception when others => raise FreeType_Exception`, please add the name of the original exception to the message. You can retrieve that with `Ada.Exceptions.Exception_Name`. This way, a user can see what exactly went wrong.
 * What's with `FT.Interfac`? I know that `interface` is a reserved name, but I do not really get why those things are not directly in `FT`. the package `Interfac` does not give any meaningful subcategory.
 * Add the copyright header from the other OpenGLAda files to your FT implementation (you can skip it for he example files). Sadly, this is rather important for any open source library.
 * Some of your subprograms return an `Errors.Error_Code` while others raise an exception. Is there a reason for this not to be consistent?

flyx:
Ah yes, and please replace `/System/Library/Fonts/Helvetica.dfont` in the example with a path to a file that is commited into the repository. Use some font with an open source license.

rogermc2:
Changed example font to Arial.ttf. Is this OK?
Since pulling your latest changes, the Open Sierra problem also occurs with Sierra!
So fixing this is my first priority.

rogermc2:
FT.Interfac originates from the original freetype_freetype I think.
My first attempt to integrate it into FT failed with lots of circular dependencies.
I've changed its name to FT.Freetype to reflect this until I can investigate incorporating it into FT.
Maybe start by moving declarations into FT?
However, I prefer to keep the functions separate if this is OK and would seem consistent with the overall structure?

rogermc2:
Copyright messages addd.

flyx:
My point about the font was that it is an absolute path to your system. Someone on Windows does not have that path in their system, nor does someone on Linux. You need to download some open source font and commit it into the repository, and then change the path to be relative to the executable pointing to that font. A good choice would probably be Google's [Noto fonts](https://www.google.com/get/noto/).

> Since pulling your latest changes, the Open Sierra problem also occurs with Sierra for this branch!

Not good, will investigate. What I changed is the time when the OpenGL dynamic functions are loaded. Previously, `GL.Init` was called from within `Glfw.Init` which did not work on Windows. The Windows documentation said that getting those function pointers only works when an OpenGL context has already been created and marked as current. Thus, I moved `GL.Init` to the constructor of a Windows (a Window includes an OpenGL context). If that now causes problems on macOS, I need to add platform-specific code. I already upgraded to High Sierra, so I'll have a look there.

> Should I raise exceptions in the API for these?

Well, if you raise the same exception for all possible errors, that would certainly be a loss of information. But you can, directly in `FT`, declare exceptions for each subcategory, like this:

    General_Error : exception;
    Glyph_Character_Error : exception;

Use `_Error` instead of `_Exception` as suffix like the Ada standard library does. Then, when you encounter an error, do:

    case Code is
       when Errors.Generic_Errors => raise General_Error with Errors.Description (Code);
       when Errors.Glyph_Character_Errors => raise Glyph_Character_Error with Errors.Description (Code);
       ...
    end case;

You can put that into an own procedure similar to `Raise_Exception_On_OpenGL_Error`.

rogermc2:
Thanks for this info, particularly regarding the error handling which I am now working on.
A weird thing happened. One of my High Sierra branches worked OK today! But the others still fail.
I was becoming suspicious of the API generation stuff but haven't been able to identify any problem.

rogermc2:
I've put a note_fonts directory in the examples directory.
I found it useful to put "file not found" into Setup_Font.
This suggests that it would be good to have this in FT.Freetype.New_Face. However, Ada.Directories is not pre elaborated so my method doesn't work.
Can you suggest a way of implementing a  "file not found" test in New_Face.
The ftgl_shader_example is again working on my Sierra machine. I've no idea what fixed it.
Yet to try with High Sierra.

flyx:
I got your code to work under macOS High Sierra without any modifications. The errors seems to be not always reproducible.

flyx:
For *file not found*, I'd just check the error code returned by FreeType. I'd assume it returns `Cannot_Open_Resource` if the file does not exist or is not readable.

flyx:
As for the package names, I think you can rename the `FT.Freetype` package to `FT.Faces` since most functions in there seem to deal with faces. Move `Init_FreeType` to `FT` and rename it to `Init` (it is obvious that it initializes Freetype as it is part of that package). Since you use the types in `FT.API` in the public interface, do not put them in `FT.API`. Put them where they belong (either in `FT` or in `FT.Faces`) and make them private. Make the package `FT.API` private since it contains stuff the user should not need to see. With the types in place, it should be possible to move all subprogram declarations from `FT.API.Freetype` to `FT.API`.

rogermc2:
OK.
Good advice. Thanks.
As long as I don't run into circular references!

flyx:
Just so that you're aware, I decided to publish v0.5 just now without the freetype stuff. The reason is that a lot has happened since the last release and I wanted to have a new one now that things are working pretty well (fixed Windows stuff earlier today).

Since this PR will still take some time, I decided to move it to the next release. Hopefully, that one will not be that far in the future :).

After your current PRs are done, I plan to move development to a `develop` branch as most git repos do. I did not do this until now because there was too little activity, but that has changed. Please wait with new PRs until I have finished with that.

rogermc2:
OK.
Sorry to not getting around to recommend routines from Common.
In particular, the projection etc. transforms.

rogermc2:
I've got most things done but can't figure out how to fix ~~~ Face_Ptr ~~~.
If I move it into ~~~ FT.Faces~~~ I get circular dependencies. If I make it private in ~~~FT.API~~~ I get a problem with ~~~FT.Faces.Face~~~ when converting it to an ~~~Object_Pointer~~~.
I have a similar problem with ~~~Glyph_Slot_Ptr~~~.

flyx:
What's the circle introduced when moving `Face_Ptr` to `FT.Faces`? It's not obvious to me.

flyx:
All occurrences I see where `FT.API.Face_Ptr` is used (e.g. `FT.Glyphs`) already depend on `FT.Faces` anyway. By removing the types from `FT.API`, you do not need to *with* it in specifications anymore and can move all `with` clauses into the bodies, breaking cycles that stem from `FT.API` needing to depend on `FT.Faces`.

rogermc2:
OK. Many thanks.
Moving the `with FT.API;` into the Faces body was the thing I missed.

rogermc2:
But,  I still have a problem with private types `Face_Ptr` and  `Glyph_Slot_Ptr` when converting to `Object_Pointer`.
For example, in `FT.Glyphs.Bitmap`, `Glyph_Pointer := To_Pointer (System.Address (Glyph_Slot))' which  complains about invalid conversion of private type.

flyx:
Do away with the conversions. `Glyph_Slot_Ptr` should be a `access Glyph_Slot_Record` in the first place. To ensure C compatibility, use `pragma Convention (C, Glyph_Slot_Ptr)`.

rogermc2:
Does this mean that I should change other system.address types to access types?
I'll try, It seems that it would be much better.
I got the original system.address bindings from the automatic GNAT binding generation tool.

rogermc2:
The High Sierra problem was caused by Apple removing Helvetica.dfont and many other dfonts!
This problem would be indicated now as we are using Error Descriptions.
It would be nice to be able to display a warning dialog for such problems.

flyx:
Typically, you use `System.Address` for C bindings only when you do not need to access the underlying object. Then it's perfectly fine to derive a private type from `System.Address` and use that.

Glad to see the High Sierra problem resolved. Since I already got your code to work on Windows 10, your code does not seem to contain anything platform-specific. We are on a good way.

rogermc2:
I've completed changing System.Address to access types wherever I found possible.

flyx:
Some minor things that could be improved:

 * Move the `Bool` subtype (back?) to `FT.API` and use the standard `Boolean` in the public interface. This is important so that people can use boolean expressions in the place of that character. You can then convert the Boolean value to `Bool` in the wrapper.
 * Get rid of `FT_` prefixes in the public API. I don't mind them in `FT.API`, but the prefix is not needed since we have packages. Prefixes that are still existing are `FT_Pos` (might be `Position`) and `FT_Vector`.
 * Hiding `FT_Vector` seems unnecessary. Make it public and do away with the `Vector_X` and `Vector_Y` functions.
 * Some other records could be made directly accessible: `FT_Bitmap_Size`, `FT_BBox` (rename to `Bounding_Box`), `Size_Metrics_Record` (rename to `Size_Metrics`). That would reduce the number of value accessor functions.

flyx:
The API is not yet memory safe – things like correctly calling `FT_Done_Glyph` and `FT_Reference_Library` / `FT_Done_Library` can be managed automatically with Ada's controlled types. Unless you want to do that yourself, tell me when you're ready with your changes so I can implement that.

rogermc2:
A couple of initial questions.
By directly accessible I assume directly accessible for read only?
I intend to provide read ability for the individual elements of a record.
Should complete records be made accessible and, if so, in what form?

rogermc2:
My understanding is that FT_Bitmap_Size, FT_BBox , Size_Metrics_Record are effectively part of the Face_Record which "belongs to" the C freetype machine and therefore should not be publicly accessible so I would prefer to keep them private.
But if you think its "safe" to make them public I'll do so.
Otherwise, I've made all the other changes.
I'd prefer for you to attend to the memory safeness.

rogermc2:
I've been keeping the other branches updated including the Geometric Algebra branch.
I've extended freetype in the String and Geometric Algebra branches to provide a more convenient AP interface and have been testing these branches at each update.
Although currently simple the Geometric Algebra use is working to my satisfaction.

flyx:
I'll have a take at memory safety, including the handling of the internal records.

It is safe to return them from a getter function because as record values, they will be copied, so the user does *not* get access to any of FreeType's internal values. I think it is nicer for the user to provide *one* getter which returns the set of values inside the record, but I will do that myself as part of the memory safety restructuring. Hold on.

rogermc2:
Ok.
Thanks, I'll be interested in your improvements.
I agree about providing one getter function. Much nicer.
Its something I've been struggling with in Geometric Algebra, in the sense of what's the best way to do it, so it will be good to see your approach.
In GA I've found the most convenient way has been to code the GA "core machine" in records (good practice I think) and provide arrays containing the data elements; but it doesn't really seem "right".

flyx:
I dug a bit into the FreeType documentation. It seems like `FT_Init_FreeType` initializes a library that should be destroyed with `FT_Done_FreeType`, not with `FT_Done_Library` like you do. But documentation is really clear, so before I change anything: Did you have another source of documentation which you got the idea to use `FT_Done_Library` from?

Background is that `FT_Library` does have reference-counting built-in, but from what I understand, one needs to use `FT_New_Library` instead of `FT_Init_FreeType` to be able to use it, which requires building a `FT_Memory` struct. This should be doable but will take some of my time.

flyx:
Well I read the source of `FT_Init_FreeType` and it calls `FT_New_Library` and does some additional initialization so it seems safe.

rogermc2:
I don't remember where I got the idea to use FT_Done_Library from; possibly from freetyp_gl.
As I tried FTGL and porting freetyp_gl before deciding to develop directly from FT, my initial design was probably based on these. I'm fairly it wasn't based directly on FT documentation.

This, from freetypt_freetype.h, seems to cover the situation:
Type
      FT_Library
Description
     A handle to a FreeType library instance.  Each library is  completely independent from the others; it is the root of a set  of objects like fonts, faces, sizes, etc.
      It also embeds a memory manager (see @FT_Memory), as well as a  scan-line converter object (see @FT_Raster).
     In multi-threaded applications it is easiest to use one   FT_Library object per thread. 
     In case this is too cumbersome,  a single FT_Library object across threads is possible also  (since FreeType version 2.5.6), as long as a mutex lock is used   around @FT_New_Face and @FT_Done_Face.                              
Note
     Library objects are normally created by @FT_Init_FreeType, and destroyed with @FT_Done_FreeType.  If you need reference-counting (cf. @FT_Reference_Library), use @FT_New_Library and  @FT_Done_Library.

rogermc2:
Just from a symmetry perspective it looks like, that as I used FT_Init_FreeType , I should have used FT_Done_FreeType, not FT_Done_Library.
If you decide that FT_New_Library and associated  FT_Memory struct should be used instead of FT_Init_FreeType to provide reference-counting then I will be happy to undertake the change if that's what's needed.

rogermc2:
I must have made a mistake way back.
I've just remembered that the original example I have been porting is from 
https://learnopengl.com/#!In-Practice/Text-Rendering
which initiates with
FT_Init_FreeType(&ft)
FT_New_Face(ft, "fonts/arial.ttf", 0, &face)
and states:
Also be sure to clear FreeType's resources once you're finished processing the glyphs:
FT_Done_Face(face);
FT_Done_FreeType(ft);
So, I don't know where the FT_Done_Library came from.

rogermc2:
I have a strange problem with the latest version.
It works on one computer but not the other.
`Hello_Character returned constraint error: raised CONSTRAINT_ERROR : erroneous memory access`
I traced the error to the 'Finalize (Object);' in 'FT.Init'.
If I remove this 'Finalize' statement the program works.
I have been unable to find a solution to this problem.

rogermc2:
I tried implementing the new API into my strings example on my "other" computer; the one that doesn't fail with  ftgl_shader_example.
However, the same error seems to be occurring.
I got a deeper error message than the one from the equivalent of ftgl_shader_example's `Setup_Font`; that is an error message from `FT.Faces.New_Face`:
`FT.Faces.New_Face error: Invalid Library Handle`

flyx:
I pushed a fix, please check if your errors go away.

I still need to migrate Face_Ptr and Glyph_Ptr to memory-managed types. My first attempt showed that this requires a lot of restructuring. Can you stop pushing commits until I'm done? Otherwise we'll be in merge hell.

rogermc2:
Great!
ftgl_shader_example now works on both my computers.
I'll try to remember not to push more commits on this branch. No way to lock out commits?
I get myself into enough merge hell as it is!
There's plenty for me to do with bringing the other branches up to date and developing the geometric algebra branch.

flyx:
Well the situation right now is that we need to get the `FT` packages in a mergeable state and that's a blocker for other things. So everything other we do *will* get merge problems as long as the `FT` API is changed, no matter how good we handle git. The only good option is to concentrate on this branch and merge it fast. I'll try to get `Face_Ptr` and `Glyph_Ptr` done today.

flyx:
I had to move around a *lot* of types to make `Face_Reference` happen. `FT.Glpyhs` is now `FT.Faces.Glyphs` because it needs a view of the `Face_Reference` internals. I moved a lot of types into the private part of `FT` so that they are visible to both `FT.Faces` and `FT.API` but not to the user. Since I had to move them around anyway, I made some records public and simplified getter functions.

However, I think I screwed up a but in `FT.Utilities.Print_Character_Metadata`. I became very confused about some of the types and when reworking the code, I think I fetched the wrong public record, but I could not find the correct one. I disabled calling the function `Print_Character_Metadata` in `Texture_Manager` so that it does not break and hope you can tell me how to fix this (or do it yourself).

Meanwhile, I saw that you declared `Character_Record` which is not a type used by the FreeType API. That also confused me a bit; why is it part of the `FT.Faces` package? It seems like a collection of some values, but I could not figure out its purpose. Are there subprogram working on it in other packages? I lost the overview. Anyway, as this is not a FreeType thing, it definitely has no place in `FT.Faces` and should be somewhere else.

flyx:
I'm done with the memory management stuff. Do I see it correctly that the Glyph stuff is currently not used?

rogermc2:
Character records are used for storing characters in the string example (which is really a bit more than an example as it provides the intended basis for general use of the system).
It originates from the https://learnopengl.com/#!In-Practice/Text-Rendering example.
The history was that I tried to develop the string example directly from this 'Text-Rendering' tutorial example but when I ran into serious difficulty I initiated this single character example as a simpler stepping stone towards the full string example.
As it turned out, when I first got this `single character` example working, I incorporated it back into the `string example` but still had difficulty so I initiated a second stepping stone which is the 'single_char_storage' branch. Finally, after getting that working I was able to build a working `string example`.
Essentialy, in the `string example`, a description of each ascii character is contained in a 'Character_Record'; the fully set of character descriptions being contained in a 'Character_Data_Vector' which is not used at all in 'ftgl_shader_example' but is essential for the 'String Example'.
My main reason for developing this system is to provide a system that easily renders strings of characters. To do this, I need to extend the FreeType API, which as far as I can understand only deals with single characters which is appropriate as it is fundamentally a font package that provides characters but, as far as I understand, no process for manipulating them. Hence the need for systems such as FTGL and FreeType_GL.
This branch, 'ftgl_shader_example', is essentally a precursor to the 'string_example' branch which, due to its origination from the  'string_example' contains a few data structures not currently utilised.
I hope this makes sense.

rogermc2:
I'm surprised if " Glyph stuff is currently not used" as it is what font characters actually are.
But its too late at night for me to look into now.
I'll try and provide an answer tomorrow.

flyx:
Okay. A good way to organise that would be to move `Character_Record` and associated subprograms into a package `FT.GL` or something so that it is clear that this is glue code for using FreeType with OpenGL. Similarity to FTGL is intentional ;).

rogermc2:
Yes, I was about to propose this.
I have a "new" package in string example which is probably where it and its  associated programs should be.
I have been struggling for a decent name for this package. 'FT.GL' sounds better than what I've currently got.
If you like you can delete 'Character_Record' and 'Character_Data_Vector' and I'll fix it up in 'string_example' tomorrow so as not to interfere with your work.
This will mean I won't need to push 'ftgl_shader_example' and cause merging problems.
Now, I really am signing off for tonight.

flyx:
Okay, that was hopefully the last hard API change I will push. I was able to move `FT.Faces.Glyphs` back to `FT.Glyphs` by introducing `Glyph_Slot_Reference` in `FT`. This make a lot of things cleaner and removes all references to `Face_Reference` from `FT.Glyphs`. I also fixed `FT.Utilities`, it does now work again with the example.

With *Glyph stuff is currently not used*, I meant that currently, the code does not use `Glyph_Ptr` (or the new `Glyph_Reference`). It only uses the `Glyph_Slot_Ptr` (which is now publicly available as `Glyph_Slot_Reference`). I asked because I implemented reference counting for `Glyph_Reference`, but this is currently untested because the code never create a new `Glyph_Reference` (it only uses the `Glyph_Slot_Reference` to draw glyphs).

flyx:
Please have a look and tell me what you think of the new interface design. I updated the example to work with the new API (at least on my machine), so you can look at how it can be used.

rogermc2:
The latest version works on both my computers.
The character_record stuff makes thing much cleaner in its new location which I've named FT.OGL; FT.GL provided some confusion with GL.
I'll have a look at the new design which I imagine will be excellent and wil inspect the glyph stuff more closely.

rogermc2:
I've checked out the new design by updating the 'single_char_storage' branch to the latest API.
No problems.
I've decided to move the fonts folder into 'freetype_gl' folder.
I haven't done it yet on the 'ftgl_shader_example' branch so as not to get into merging difficulties with your development.

rogermc2:
I've integrated the new design into the 'string_example' branch.
I found the need for two additional functions, 'FT.Glyphs.Bitmap_Width' and  'FT.Glyphs.Bitmap_Rows'.
I've implemented them in the 'string_example' branch which now works fine.
I haven't implemented them in 'ftgl_shader_example' yet per your request not to commit.
Is it OK to implement them now?

rogermc2:
With regard to exception handlers, I was thinking that many can be removed prior to merging as, from my perspective, they performed a valuable service for me during development but Ince the design has stabilised and working well there is really no need for them.
So I am suggesting that, just prior to merging, I go through the code and remove many of them.

flyx:
Yes, removing those *catch-all-and-raise-new-exception* handlers would be a good thing. Exceptions that are raised based on the error code should obviously stay.

Another good thing to do would be to go over all subprograms that call API functions and ensure that they do the appropriate `null` check before calling the API function because the FreeType implementation never checks for null, but instead just dereferences the pointers it gets.

After those things are done, I think this PR is ripe for merging.

flyx:
Why do you need `Bitmap_Width` and `Bitmap_Rows`? `Bitmap_Record` is public, so `FT.Glyphs.Bitmap (My_Glyph_Slot).Width` would work, wouldn't it?

flyx:
I renamed `FT.Faces.Slot` to `FT.Faces.Glyph_Slot`, I think that's more clear, and I removed unnecessary exception handlers from the `FT` packages (there weren't many; those in the example can stay). I reviewed memory management, found and fixed a leak.

We now have this in a state that could be merged unless you have further propositions. There is of course the option to clean up `FT` from any `GL` references and make it a standalone library so that it can be used for other purposes. Basically, that would mean we remove all `GL.Types` types and replace them with self-defined types or those from `Interfaces.C`. We'd then have a project `freetype` which does *not* depend on `opengl`, and separately, `freetype-opengl`, which adds glue to OpenGL (such as loading a texture from a bitmap). But I think this is a transformation we can tackle later.

flyx:
Oh well, I got rid of `GL.Types` everywhere in `FT`. The only remaining reference to `GL` is the `Image_Source`. We will get rid of that later, when we merge `FT.OGL`. We can then put a transformation function there.

I fixed some small errors in record representation. Also, I renamed `Errors` to `FT.Errors` (it was originally there, wasn't it?). We cannot have a global package `Errors`, that is likely to name-clash with a package from a user's code.

rogermc2:
I'm glad that GL references are going.  This is a bother that I have been having with Geometric_Algebra trying to keep references to GL to a minimum so that the non-graphics code is straight Ada. Your getting rid of GL here gives me more confidence snd encouragement to do the same with GA.
While `Bitmap_Record ` contains`Bitmap_Width` and `Bitmap_Row` it does not contain  `Bitmap_Top` and `Bitmap_Left` which are in `Bitmap's` parent and which I use function calls for.
So this is why I thought functions `Width` and `Height` were missing.
I'll try to use  `Bitmap_Record ` for `Width` and `Height`. My current quick attempt is failing, but that's probably just me. It should work.

rogermc2:
Regarding GL.
In my Geometric Algebra development, should I minimise my use of GL and, where I do need to use it, type cast results to core Ada?
This seems to me a good way to go.

rogermc2:
I'm using Bitmap_Record for Width and Height OK so no need for  `Bitmap_Width` and `Bitmap_Rows` getters.
I've added `Glyph_Slot` checks to FT.Glyphs. Hopefully the right thing to do?
I can't think of anything else that needs to be done.

flyx:
Checking for `Glyph_Slot_Reference.Data` not being null is not necessary, it is declared as `not null Glyph_Slot_Ptr`. We only need the check this in `FT.Faces.Glyph_Slot`, afterwards, it cannot be null. Please revert that commit.

rogermc2:
OK. Reverted.

flyx:
You don't seem to have pushed that revert commit

rogermc2:
Should be OK now.
I don't understand what happened.
git status showed nothing to commit so I thought it happened automatically.
However, following your message, I pushed and the revert seems to have gone through.

flyx:
This was a lot of work, but I think it was worth it. Thanks! I'll look forward to your other PRs.

rogermc2:
Likewise; your invaluable assistance is greatly appreciated.

PR #90, opened 2017-09-22
Single char storage

rogermc2:
This example tests/illustrates the use of Texture_2D.Storage and Texture_2D.Load_Sub_Image_From_Data for a single character as a prelude to effectively implementing the rendering of strings which, so far, has proved problematic.


flyx:
The `.gitignore` file starts getting too large for my taste and causes frequent merge problems (as seen here). I think it would be better to keep the names of the example executables out of the main `.gitignore` file and instead have one `.gitignore` file per example directory that simply specifies the name of the executable generated there. Can you do this starting with this example? I will take care of the existing lines in the main `.gitignore` myself.

rogermc2:
OK.
I've add .gitignore to all freetype_gl branch directories.

rogermc2:
Development frozen.
No further changes planned.
Please review.

rogermc2:
Should close until first freetype example completed.

PR #94, opened 2017-10-04
Initial commit of mesh-view example

leo-brewin:
I've created a branch that contains just the addition of the mesh-view example. I'm not sure if I've followed correct Git workflow and I'm happy to take any advice (including starting from scratch :).


rogermc2:
Thanks to Google, I now have mesh-view branch and have been able to build and run meshview.
I'm pretty sure that one of the first things that Felix will want is for your common to be merged with the existing common. I presume that he will want at least that done before he merges the PR.
Before I can do anything with your code I need to find out how to synchronize our two branches.

rogermc2:
I've just changed all common files, except the initialize files, to use examples/common files and all works well. 
Of course these changes are only on my computer.
Just looking at your Initialize, I should take Glfw.Input.Mouse.Cursor_Mode  and Show_GL_Data out of my version as they should be left up to the user's main initialization program.

rogermc2:
Just a small point; we are now including file names in error messages.
E.g `"An exception occurred in meshview.meshview."`

rogermc2:
I'm wondering if you keyboard and mouse callbacks could be moved to common/controls as they appear to be of general use.

rogermc2:
I suspect libobj could also be move to examples/common as it presumably is generally useful.
Did you consider using streams (if appropriate)? Felix recommended streams to me when I was developing 'load_dds` and I found them a great improvement over what I had originally.

rogermc2:
Felix doesn't like commented out code as it can be a distraction to users of the code.  Commented out code should be removed.

flyx:
@rogermc2 Actually, I am not that fond of exception messages including package / subprogram names. I just kept them in your PR because there were more important things to tackle. There's actually no need to include the subprogram's name in the exception message because locating an exception's origin is already [a feature of GNAT][1]. It can easily be used with the visual debugger in GPS and in addition to the exception source line, it provides a complete stack trace. I have no strong opinion on it, but this is definitely not a prerequisite to get a PR merged ;).

It may take a while for me to review this, since current focus is merging #87 .

 [1]: http://www.adacore.com/adaanswers/gems/gem-142-exceptions/

rogermc2:
With regard to Initialize, I now recall that my original idea with respect to such overridable things as `cursor` was that I would initialize to a default setting and that the user could override them in their own application specific "setup" routine, which I think is what I've done in at least one example.
My current thinking is that I'll remove `Show_GL_Data` from `Initialize` and, if agreed, add  `Class` to the `Main_Window` declaration.
Although these are only small changes I intend to create a separate branch `update_intialize` to implement them because of their potential impact on many examples.
For example I would like to include  `Show_GL_Data` in basic examples.
Also, if `Class` is implemented, then all examples will need change.
Does this seem OK?

leo-brewin:
Greetings Felix and Roger,

I'm in no particular rush to make any changes at the moment -- I'm more
than happy to wait for Felix to review the PR before I do anything. But I
do agree with Roger that it's not sensible to have two separate "common"
directories. At some point these should be consolidated into one (with the
required changes made to all the other examples).

Cheers,
Leo

flyx:
So I finally can have a look at this. I won't read all the code, but there are some things I stumbled upon:

 * By any chance, can the included PNG library be replaced by using the SOIL binding? Problem is, that PNG library is licensed under GPL which I'd rather avoid. Thing is, it is not directly harmful here (since it is not part of the main sources), but when a potential user scans over the repository for GPL'd stuff, they might be scared away.
 * There is a strange binary file named `ZLib for Ada thick binding. download | SourceForge.net.webloc` included. Looks like garbage, is it required?

leo-brewin:
I'm pretty well flat out with other things too so I'm happy with the delay in processing the pull request :)

- I use the PNG library only as a way to map (u,v) coordinates to rgb colours. I use the colour map to help visualise scalar data over a 2d-grid (e.g., temperature plotted as height and colour over the xy-plane). If I can do the same in SOIL then great. I'm not familiar with SOIL. How hard would it be to load an image file then get it's rgb values for each pixel?
- The Zlib library is required by the PNG library. The file you noted is just a URL to the Zlib home page. It can be deleted.

flyx:
I am very sorry, I have left this lying around for a looong time.

I just tried to build and execute your code. Right after running, a window pops up and closes again, then I get

    raised PROGRAM_ERROR : EXCEPTION_ACCESS_VIOLATION

Any idea what might cause this? What operating systems have you tested this with?

leo-brewin:
Hi Felix,

No problems. I too have had a lot on my plate (and sorry about my slow repose to you :).

I'm on a Mac running macOS High Sierra 10.13.3. The glfw and glew libs were installed using the "brew" package manager for macOS. The versions are: glfw-3.2.1 and glew-2.1.0. I have recompiled the OpenGLAda and mesh view codes and they both run fine on my mac. I don't know what the problem could be. 

The only libs that I use that aren't used anywhere else in OpenGLAda are the png and zlib libraries. The libPNG is used in only routine, libobj.adb. The declare block in lines 252-274 serve to set the colours for each vertex in the mesh. That block could be changed to set the colours to a single shade of grey (just for testing). If the code then runs then it's clear that the access violation has something to do with the png/zlib libs. I know I'm plucking at straws. But given that the rest of OpenGLAda runs fine I suspect the problem must be in the interaction between OpenGLAda and the png/zlib's.

Cheers,
Leo

PR #96, opened 2017-10-05
Single char storage

rogermc2:
I've fixed text_manager.
Could be other problems that I can't find.


rogermc2:
It looks like I've messed up my master!
Can I recover???

rogermc2:
I may have (hopefully) fixed my master.

src/gl/implementation/gl-objects.adb
```diff
@@ -43,17 +43,7 @@ package body GL.Objects is
...
-               exception
-                  when others =>
-                     --  cannot let this escape as we're in a Finalize call and
-                     --  thus that error cannot be properly catched. Chances are
-                     --  that if the destructor fails, the context already has
-                     --  vanished and thus we do not need to worry about
-                     --  anything.
-                     null;
-               end;
+               Reference.Destructor (Reference);
```
flyx: Now that's here again ;).

src/gl/interface/gl-pixels.ads
```diff
@@ -345,6 +346,9 @@ private
    for Data_Format use (Stencil_Index   => 16#1901#,
                         Depth_Component => 16#1902#,
                         Red             => 16#1903#,
+                        Green           => 16#1904#,
+                        Blue            => 16#1905#,
+                        Alpha           => 16#1906#,
```
flyx: What are those doing here? This enum is for the `glTexImage?D` functions. According to the [spec](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexImage2D.xhtml), they don't take `GL_GREEN`, `GL_BLUE` or `GL_ALPHA`.

rogermc2: OK.
I'll remove them.

src/gl/implementation/gl-objects.adb
rogermc2: I looked for this but couldn't find the previous comment.
I'll try and fix it yet again.

flyx:
What exactly is the difference between this example and the existing one? Using `Texture_2D.Load_Sub_Image_From_Data` instead of `Texture_2D.Load_Image_From_Data`?

I think these two examples need more description to clarify what they are about. For me, it is not clear why `Load_Sub_Image_From_Data` is now used and what the consequences are compared to `Load_Image_From_Data`.

rogermc2:
On reviewing the code, I see what you mean.
As  a development example, it was essential for me in sorting out how to use `Texture_2D.Storage` and `Texture_2D.Load_Sub_Image_From_Data`.
Finding the correct data to supply to each of the parameters for these procedures was, I remember, the big difficulty.
However, as a general example, its not very useful unless I provide an intermediate texture storage step between `Texture_2D.Storage` and `Texture_2D.Load_Sub_Image_From_Data`.
But this is probably not useful as I expect the complete string example is just as informative.
So I'll withdraw this PR and submit the complete string example.

PR #97, opened 2017-10-07
String example

rogermc2:
This provides the final step in this journey of providing an Ada FreeType interface for rendering strings.


src/freetype/ft-ogl.adb
```diff
@@ -0,0 +1,285 @@
...
+   use Interfaces.C;
+    procedure Load_Vertex_Buffer is new
+     GL.Objects.Buffers.Load_To_Buffer (GL.Types.Singles.Vector4_Pointers);
+
+   type Character_Data_Vector is array (Natural range <>) of Character_Record;
+
+   Face_Ptr             : FT.Faces.Face_Reference;
+   Vertex_Array         : GL.Objects.Vertex_Arrays.Vertex_Array_Object;
+   Vertex_Buffer        : GL.Objects.Buffers.Buffer;
+   Extended_Ascii_Data  : Character_Data_Vector (0 .. 255);
```
flyx: I think these global vars should be in a private type `Font_Renderer` instead. Maybe together with the `Face_Reference` above. So the user initialises a `Font_Renderer` object and afterwards can draw text with it.

This will also allow loading multiple fonts at the same time.

src/freetype/ft-ogl.adb
```diff
@@ -0,0 +1,285 @@
...
+      --  2D quad as two triangles requires 2 * 3 vertices of 4 floats
+      Vertex_Data    : Singles.Vector4_Array (1 .. Num_Vertices);
+   begin
+      --  Blending allows a fragment colour's alpha value to control the resulting
+      --  colour which will be transparent for all the glyph's background colours and
+      --  non-transparent for the actual character pixels.
+      GL.Toggles.Enable (GL.Toggles.Blend);
+      GL.Blending.Set_Blend_Func (GL.Blending.Src_Alpha,
+                                  GL.Blending.One_Minus_Src_Alpha);
+      GL.Objects.Programs.Use_Program (Render_Program);
```
flyx: Since the shaders are not part of this library (only provided for examples), I don't think it is a good idea to have program setup code in here. Users should setup an appropriate program for themselves. OTOH, if the exact shaders in your example are necessary for this package to do anything useful, they should be part of the main sources instead of the examples. Can you give a pointer about how vital the correct shaders are for this to work, and how much sense it would make using user-provided shaders?

src/freetype/ft-ogl.adb
```diff
@@ -0,0 +1,285 @@
...
+      GL.Blending.Set_Blend_Func (GL.Blending.Src_Alpha,
+                                  GL.Blending.One_Minus_Src_Alpha);
+
+      Vertex_Array.Initialize_Id;
+      Vertex_Array.Bind;
+
+      Vertex_Buffer.Initialize_Id;
+      Array_Buffer.Bind (Vertex_Buffer);
+
+      for index in Extended_Ascii_Data'Range loop
```
flyx: On the long run, we'll want to be able to print more than just ANSI characters. Therefore, I wonder if this approach of preloading all character images is wise. How about this alternative approach:

Provide a type `Renderable_Text`. This can be created with a `Text_Renderer` from a String and is used as input for `Render_Text`. When creating `Renderable_Text`, you check whether each character is loaded and if not, load it. `Renderable_Text` then only contains an index list pointing to the characters used, and when you render it, you can be sure that the characters it uses are properly loaded.

src/freetype/ft-ogl.ads
```diff
@@ -0,0 +1,44 @@
...
+-- OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+--------------------------------------------------------------------------------
+
+with GL.Objects.Programs;
+with GL.Types;
+with GL.Types.Colors;
+with GL.Uniforms;
+
+package FT.OGL is
+   type Character_Record is private;
```
flyx: There is no way of getting access to a `Character_Record` value with the public subprograms in this package, so there is no need to expose this declaration.

src/freetype/ft-ogl.adb
```diff
@@ -0,0 +1,285 @@
...
+-- THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+-- WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+-- MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+-- ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+-- WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+-- ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+-- OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+--------------------------------------------------------------------------------
+
+with Ada.Text_IO; use Ada.Text_IO;
```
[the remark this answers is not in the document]
rogermc2: single_storage example, errors.ads and errors.adb removed
I/O operations removed from OGL

src/freetype/ft-ogl.adb
```diff
@@ -0,0 +1,285 @@
...
+with FT.Glyphs;
+
+package body FT.OGL is
+   use Interfaces.C;
+    procedure Load_Vertex_Buffer is new
+     GL.Objects.Buffers.Load_To_Buffer (GL.Types.Singles.Vector4_Pointers);
+
+   type Character_Data_Vector is array (Natural range <>) of Character_Record;
+
+   Face_Ptr             : FT.Faces.Face_Reference;
```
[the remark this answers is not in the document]
rogermc2: Fixed.

src/freetype/ft-ogl.adb
rogermc2: This sounds an excellent idea.
However, I would prefer to develop it as a separate example if that's OK.
I prefer to preserve the original context of this example to keep it in line with the original C example that it is based on.

src/freetype/ft-ogl.adb
rogermc2: I had been thinking about moving the shaders and associated setup operations into OGL so that, for rendering text, the user would not have to know or do anything about the situation.
However, a problem is the the shaders are machine dependent in that the GLSL version number appropriate to the computer that they are being used on.
My feeling is that these shaders are specific and to OGL and the best thing would be that the user did not have to provide them.
Although of given this some thought, I haven't come to any conclusion.
With regard to the blending function, I found it essential to set and reset it for text rendering use as if permanently set it causes flickering problems with other graphics.
However, I've just realised that I should record the blend state on entry and return it to that state before leaving the routine.
More thought needed on my part.
With this I'll appreciate any suggestions you can provide.

src/freetype/ft-ogl.adb
rogermc2: Preloading all character images originates from the original C source of this example, the context of which I would like to preserve.
I suggest making an additional example incorporating all of your ideas which certainly should be incorporated into the final "production" version of OGL.
As the current OGL falls well short of the ideal version that you have described, I'm wondering if OGL should be moved out of the current `freetype` library and back into the example itself until it is improved to its most useful implementation?

src/freetype/ft-ogl.ads
rogermc2: Fixed.

rogermc2:
I think I've fixed everything the needs fixing.
I am currently thinking about whether or not the shaders in your example are necessary for this package to do anything useful.
My current thinking is that they are and therefore should be part of the main sources instead of the examples.  However, there is the problem of machine dependence to address?

With regard to printing more than just ANSI characters I really think that this should definitely be catered for but should be the province of another example.
Ideas for the complete implementation of a widely useful string rendering system are pretty much outside the domain of my knowledge. So, I must rely on your recommendations for ways of extending this example.

rogermc2:
On further thought, as OGL doesn't appear to be as generally useful as I thought it might be, it might be better to move  OGL back  into the `string example` (where it originally came from) and develop a more general version independently?

rogermc2:
I've made an attempt to transfer the shader processing into OGL in my extensions branch.
However, a problem arises that the examples/common files `Maths' and 'Program_Loader' are required which, obviously, are not part of OpenGLAda's library and, so, don't seem to be visible to OGL (which presumably is a good thing from the OpenGLAda library point of view.).

flyx:
If this example is important because it mirrors what one would typically do in C, then I agree that the stuff currently in OGL should be moved into the example. FT.OGL should provide a simplified interface with only the necessary things visible to the user.

About machine dependence of the shaders: Is there anything that forbids us to just use the oldest possible shader version? Wouldn't that work on all modern machines?

rogermc2:
OK, I'll move it back into the example and also check out how far back I can set the shader version.
Deciding what "a simplified interface with only the necessary things visible to the user" is a bit of a challenge.
What I currently have is:
~~~
procedure Initialize_Font_Data (Font_File : String;
                                   Projection_Matrix : GL.Types.Singles.Matrix4);
procedure Render_Text (Text   : String; X, Y, Scale : GL.Types.Single;
                          Colour : GL.Types.Colors.Basic_Color;
                          Projection_Matrix : GL.Types.Singles.Matrix4);
~~~
Would this make sense as a simple user interface?

flyx:
I think a simple user interface would render the text onto a texture so that the user can then place it as desired. This would remove most of the `Render_Text` variables. Additionally, it would have the `Font_Renderer` I described above. It would also provide a way to query how long a String would be if rendered so that the user can decide if and where to split it into multiple lines.

rogermc2:
What you describe seems admirable.
However, its a bit beyond my knowledge of graphics in general so I'm not sure how I should proceed but I'll study your comments and try to come to a better understanding of the situation.
In the meantime I'll first make the changes to string_example to make it a normal stand-alone example that uses our existing freetype library without any extensions.

rogermc2:
I've moved OGL back into the example but have a problem accessing `Bitmap.Width` and `Bitmap.Rows`.
For example 
```
Width := GL.Types.Int (Glyph_Slot.Data.Bitmap.Width);
~~~
which previously worked now produces  `invalid prefix in selected component "Glyph_Slot"`
with `Glyph_Slot` declared as 
~~~
Glyph_Slot     : constant FT.Glyph_Slot_Reference := Face_Ptr.Glyph_Slot;
~~~
other references to `Glyph_Slot` present no problems.

rogermc2:
I moved OGL into the string example directory, but compilation fails as I can't seem to access 
`Glyph_Slot.Data` as
~~~
Width := GL.Types.Int (Glyph_Slot.Data.Bitmap.Width);
~~~
and `Rows` fails with invalid prefix in selected component "Glyph_Slot"
 `invalid prefix in selected component "Glyph_Slot"`.
Getter functions would be nice as they would make the code more consistent wth the `Bitmap_Left` and 'Bitmap_Top` function calls.

rogermc2:
GLSL version 400 works
GLSL version 330 works
Version 300 - 320:
~~~
ERROR: 0:1: '' :  version '300' is not supported
ERROR: 0:2: '' :  #version required and missing.
ERROR: 0:3: '0' : syntax error: integers in layouts require GLSL 140 or later
~~~

flyx:
Instead of

    Glyph_Slot.Data.Bitmap.Width

do

    FT.Glyphs.Bitmap (Glyph_Slot).Width

You cannot use prefix notation for accessing the `Bitmap` function because `Glyph_Slot_Reference` is not a tagged type.

rogermc2:
OK, thanks.
That problem fixed but not quite there yet.

rogermc2:
String rendering code moved from freetype library back into text_rendering.
Ready for review again.

rogermc2:
OK. I'll sought it out.
I don't know how those other files got in there and I thought I'd changed the file name!!!!
 I did change the file name!  I just forgot to push the changes.
Sorry about that, fixed now. My brain must have gone into overload mode!
With regard to the unnecessary files, could this have something to do with my changes to `.gitignore` and the new `.gitignore` in `examples/freetype_gl`?

rogermc2:
I've removed the  unnecessary files and updated .gitignore to ignore `examples/superbible/simple_texture/simple_texturing`.

flyx:
Can you rename `FT.OGL` so that there will be no name conflict when we add the *real* `FT.OGL`?

rogermc2:
I had renamed them but the originals somehow remained.
I've deleted them now.

flyx:
Do you have other unmerged examples using the FreeType stuff? otherwise, I would like to re-organize the project structure a bit.

rogermc2:
None.
Go ahead and reorganize.
Can you make the user interface as simple as possible?
My original concept was that the user should do everything via `Face_Ptr`, now  `Face_Reference` I suppose. But I don't know that this is possible or advisable now.

rogermc2:
I am using an earlier version on another project.
Should I wait until your reorganization is complete before switching to the master version?

flyx:
I made the freetype stuff independent from OpenGLAda and published it separately at https://github.com/flyx/FreeTypeAda . It is still part of this repository, I use [git-subrepo][1] to keep it in sync (much nicer than the standard `git submodule`). I also added style checks to the compiler to enforce proper coding style. And lastly, I made the build fail on warnings. This is to force us or any other contributor to commit sane code.

Now I'll see what I can do for a nicer FreeType interface.

 [1]: https://github.com/ingydotnet/git-subrepo

rogermc2:
Sounds good.
The worthwhile adding of style checks has shown that some changes are needed in some examples.
Should I open a new branch to fix these?
Presumably I shouldn't do it in master!

flyx:
Yes, you can fix them on a new branch. You can branch from master; I won't do more reorganising work to the structure for now. A good thing to do would be to make all examples use the same compiler settings as the main project. I did this with the FreeType examples if you want to have a look.

Btw, I removed your `-L/usr/local/lib` from the freetype project files. If you need it, do `export LIBRARY_PATH=/usr/local/lib` before compiling.

rogermc2:
I am getting this error in all examples:
~~~
An exception occurred in Initialize.
An exception occurred in Hello_Character.
raised PROGRAM_ERROR : glfw-windows.adb:172 accessibility check failed
~~~

flyx:
Sorry, my bad. Fixed it.

rogermc2:
Thanks, working again.

PR #100, opened 2017-10-30
String example extensions

rogermc2:
Example using GL.Text
Common made into library.


examples/common/src/main.adb
```diff
@@ -0,0 +1,6 @@
+procedure Main is
```
flyx: What is this file for? You should declare `common` as `library project` instead so that it does not need a `main`.

rogermc2: It shouldn't be there. I'm removing it.
It must have got automatically generated by GPS.
Still, I should have noticed it.

rogermc2: `main` deleted.
Bad styles corrected.

rogermc2:
Thanks Flyx.
This marks a milestone in my development.

rogermc2:
Thanks Flyx.
This marks a milestone for me on my OpenGL experience.
However, I'm wondering about the use of this develop branch.
At first I didn't relalize and merged upstream/master which of course caused problems.
When I merged upstream/master things seem to have recovered.

rogermc2:
There were problems due to removal of opengl shared.
I have submitted a separate PR to fix those.

rogermc2:
You did mention that you intended to use a  `develop branch`.
Does this mean that I should create new branches using
`git checkout -b <new branch> upstream develop`?

flyx:
I believe the correct syntax is `git checkout -b <new branch> upstream/develop` but yes.

rogermc2:
That's what I meant. Typo. Sorry and thanks.

PR #101, opened 2017-11-07
Update common.gpr

rogermc2:
Remove references to OpenGL.Shared.


examples/freetype_gl/text_rendering_2/text_rendering_2.gpr
```diff
@@ -10,7 +9,9 @@ project Text_Rendering_2 is
    for Exec_Dir use ".";
    for Main use ("render_text.adb");
 
-   package Compiler renames Opengl_Shared.Compiler;
+package Compiler is
+   for Default_Switches ("ada") use ("-gnatyctux");
+end Compiler;
```
flyx: Why does this not rename `OpenGL.Compiler` now?

rogermc2: I tried  `package Compiler renames OpenGL.Compiler;`
but the Builder fails with `"OpenGL" is not an imported or extended project`.
If you can help me fix it, should I should I change all examples to use it?

flyx: Well you need a

    with "../../../opengl.gpr";

Unlike in Ada, the parent project is not automatically imported when you *with* `opengl-text.gpr`.

rogermc2: OK. That works
I thought I tried that bit i'd copied other `limited with` lines which was my problem.
Thanks for the fix.
I'd like to use it in all examples.
Will this be OK.

flyx: Sure.

rogermc2:
OK. Please hold off merging until I've done that, probably, tomorrow.

rogermc2:
I have come across a problem with the Durian example `buffers_and_textures` which involves SOIL.
Initially the problem was missing `stdio.h` which I overcame by installing X-Code command tools.
Apparently they get deleted every time that you upgrade XCode!
I recall that last time I "fixed" this problem I used an alias to a version of `std.io` in `usr` or `usr/local`.
This time installing X-Code command tools fixed that but now the build fails with:
~~~
Builder results (4 items)
   Authorization.h (1 item)
       193:7 error: variably modified 'bytes' at file scope
   CGFont.h (1 item)
  CGPath.h (2 items
~~~
Sounds like possibly wrong C compiler?
Can you suggest a solution. Thanks.

rogermc2:
Style checking updates completed.

rogermc2:
Style corrections have been completed.

PR #116, opened 2018-03-07
Draw example

rogermc2:
This demonstrates OpenGL draw commands.
It includes some additions and corrections to OpenGLAda including:
Implementing Draw_Arrays_Instanced.
Correcting GL.Objects.Buffers.Size and Usage enum values (according to the gpl3 header in my OpenGL framework)
Correcting the size parameter of the call to API.Buffer_Sub_Data in Set_Sub_Data.

Some new procedures have been added to Common/Utilities.


rogermc2:
The real reason for this example is that I need it as a precursor to my `transform feedback` attempt.

src/gl/specs/gl-api.spec
```diff
@@ -218,6 +218,10 @@ spec GL.API is
                           First : Int; Count : Size) with
      Static  => "glDrawArrays",
      Wrapper => "GL.Objects.Vertex_Arrays.Draw_Arrays";
+   procedure Draw_Arrays_Instanced (Mode  : Connection_Mode;
+     First : Int; Count : Size; Instances : Size) with
```
flyx: Is there a particular reason `First` is of type `Int` and not `Size` here? It's `Size` in the public interface and does not seem to be able to get negative.

rogermc2:
I'll look into this and fix it.
As you say I should be able to change it to `Size`

rogermc2:
Added fix to Maths.Scaling_Matrix (Scale_Factor : Single)

PR #119, opened 2018-03-15
Indexing example

rogermc2:
This example demonstrates indexing of vertex buffers via element buffers


flyx:
I think at some point in time, we should split the examples into an own repository, as the number of binary files and files in general gets bigger and bigger.

rogermc2:
OK.
Currently I have three more in the queue waiting to PR.
Hopefully I shouldn't need any more for a while.
New examples are prompted by new features that I have problems with in the main Ada project that I am working on.

PR #120, opened 2018-03-16
New implementations

rogermc2:
Additional GL implementations that I need for a current project.
Recommended changes will be appreciated.


src/gl/interface/gl-objects-buffers.ads
```diff
@@ -13,10 +14,14 @@ package GL.Objects.Buffers is
                          Dynamic_Draw, Dynamic_Read, Dynamic_Copy);
 
    type Buffer_Target (<>) is tagged limited private;
+   type Texture_Buffer_Target (<>) is tagged limited private;
```
flyx: Inherit from `Buffer_Target` and move this type below all `Buffer_Target` subroutines to inherit the subroutines of `Buffer_Target`. That will get you rid of the duplicates.

src/gl/implementation/gl-objects-programs.adb
```diff
@@ -150,26 +151,77 @@ package body GL.Objects.Programs is
...
    end Tess_Gen_Vertex_Order;
 
-   function Transform_Feedback_Buffer_Mode (Subject : Program)
+   function Transform_Feedback_Buffer_Mode (Object : Program)
                                             return Buffer_Mode is
       function To_Buffer_Mode is new Ada.Unchecked_Conversion (Int, Buffer_Mode);
    begin
       return To_Buffer_Mode
-        (Program_Int_Param (Subject, Enums.Transform_Feedback_Buffer_Mode));
+        (Program_Int_Param (Object, Enums.Transform_Feedback_Buffer_Mode));
```
flyx: The usual way to do this is to define `Program_Buffer_Mode_Param` and import that directly with the differently typed return value. Please do that instead of using `Ada.Unchecked_Conversion`.

rogermc2:
Thanks for your instructions.
It might take me a day or two to investigate and fix things.

src/gl/implementation/gl-objects-buffers.adb
```diff
@@ -47,6 +46,41 @@ package body GL.Objects.Buffers is
...
+        then
+         API.Bind_Buffer (Target.Kind, Object.Reference.GL_Id);
+         Raise_Exception_On_OpenGL_Error;
+         if Cursor = Buffer_Maps.No_Element then
+            Current_Buffers.Insert (Target.Kind, Object);
+         else
+            Current_Buffers.Replace_Element (Cursor, Object);
+         end if;
+      end if;
+   end Bind;
```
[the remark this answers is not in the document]
rogermc2: A quick look shows this one is for `Target : Texture_Buffer_Target` whereas the previous one is for `Target : Buffer_Target`.
Can I get the previous one to accept  `Texture_Buffer_Target` ?

src/gl/implementation/gl-objects-buffers.adb
```diff
@@ -69,6 +103,17 @@ package body GL.Objects.Buffers is
...
+   procedure Load_To_Texture_Buffer (Target : Texture_Buffer_Target;
+                                     Data   : Pointers.Element_Array;
+                                     Usage  : Buffer_Usage) is
+      use type C.long;
+   begin
+      API.Buffer_Data (Target.Kind,
+        Pointers.Element'Size * Data'Length / System.Storage_Unit,
+        Data (Data'First)'Address, Usage);
+      Raise_Exception_On_OpenGL_Error;
+   end Load_To_Texture_Buffer;
```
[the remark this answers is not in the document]
rogermc2: Sama as above.
This is for `Target : Texture_Buffer_Target` whereas the previous one is for `Target : Buffer_Target`.

flyx:
I updated the specs because `glDrawArraysInstanced` needed to be loaded dynamically. You should merge `gl-api.spec` into your branch and regenerate the generated files.

By the way, I commited some stuff that allows me to build a Windows installer for OpenGLAda :). I'd like to make a new release shortly and provide that installer, so that Windows folks can use this library more easily.

rogermc2:
I hope that I've merged `gl-api.spec` into this branch and regenerated files correctly.
I'm never 100% confident with merging.

src/gl/implementation/gl-objects-buffers.adb
```diff
@@ -92,6 +153,23 @@ package body GL.Objects.Buffers is
...
+                                      Index_Type : Unsigned_Numeric_Type;
+                                      Element_Offset : UInt := 0;
+                                      Instance_Count : UInt := 0) is
+      Element_Bytes : UInt;
+   begin
+      case Index_Type is
+         when UByte_Type => Element_Bytes := 1;
+         when UShort_Type => Element_Bytes := 2;
+         when UInt_Type => Element_Bytes := 4;
+      end case;
```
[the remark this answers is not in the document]
rogermc2: Have done.

src/gl/implementation/gl-objects-programs.adb
```diff
@@ -150,26 +151,77 @@ package body GL.Objects.Programs is
...
+      use Interfaces.C.Strings;
+      String_Size : constant Natural := Varyings'Length;
+      Num_Strings : constant GL.Types.Int :=
+        GL.Types.Int (Count (Varyings, ",") + 1);
+      Prev        : Natural := 1;
+      Vary_Index  : size_t := 1;
+      C_Varyings  : chars_ptr_array (1 .. Interfaces.C.size_t (String_Size));
+   begin
+      if Num_Strings = 1 then
+         C_Varyings (1) := New_String (Varyings);
```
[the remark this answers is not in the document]
rogermc2: I'm struggling with this.
I can produce a "string" of null terminated strings but I don't understand how to "hand out pointers to the characters".
I presume these pointers are what go into the chars_ptr_array required by the `glTransformFeedbackVaryings` interface but I can't figure out how to get the pointers.
From your comments, I don't suppose the  `Interfaces.C.To_C` functions can be used?

src/gl/interface/gl-objects-buffers.ads
rogermc2: Have done (hopefully correctly).
Duplicates deleted

src/gl/implementation/gl-objects-buffers.adb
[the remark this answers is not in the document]
rogermc2: Deleted.

src/gl/implementation/gl-objects-buffers.adb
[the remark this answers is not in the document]
rogermc2: Deleted

src/gl/implementation/gl-objects-programs.adb
rogermc2: Have done (hopefully correctly)

src/gl/interface/gl-objects-buffers.ads
```diff
@@ -57,17 +60,25 @@ package GL.Objects.Buffers is
...
-
+   procedure Draw_Elements_Instanced (Mode : Connection_Mode; Count : UInt;
+                                      Index_Type : Unsigned_Numeric_Type;
+                                      Element_Offset : UInt := 0;
+                                      Instance_Count : UInt := 0);
    procedure Invalidate_Data (Object : in out Buffer);
    procedure Invalidate_Sub_Data (Object : in out Buffer;
                                   Offset, Length : Long_Size);
 
+   subtype Texture_Buffer_Target is Buffer_Target;
```
flyx: Please make this a derived type, not a subtype. Only a new type will disallow calling `Allocate` below with a `Buffer_Target`.

src/gl/specs/gl-api.spec
```diff
@@ -227,6 +227,12 @@ spec GL.API is
                             Index_Type : Unsigned_Numeric_Type;
                             Indices    : Low_Level.IntPtr) with
      Static => "glDrawElements", Wrapper => "GL.Objects.Buffers.Draw_Elements";
+   procedure Draw_Elements_Instanced (Mode       : Connection_Mode;
+                                      Count      : Size;
+                                      Index_Type : Unsigned_Numeric_Type;
+                                      Indices    : Low_Level.IntPtr;
+                                      Instances  : Size) with
+    Static => "glDrawElementsInstanced", Wrapper => "GL.Objects.Buffers.Draw_Elements_Instanced";
```
flyx: This needs to be `Dynamic`, not `Static`.

src/gl/interface/gl-objects-buffers.ads
rogermc2: According to my reference manual this should work:
~~~
 type Texture_Buffer_Target is limited new Buffer_Target with null record;
~~~
but it produces an error message:
~~~
completion of nonlimited type  cannot be limited
~~~
What am I missing?

rogermc2: Fixed

src/gl/specs/gl-api.spec
rogermc2: Fixed

rogermc2:
I think of fixed everything except `Transform_Feedback_Varyings` for which I have not found a suitable implementation.
Please provide details on how to "hand out pointers to the characters there" which I don't understand and, after many hours, haven't been able how to figure out how to implement.
API.Transform_Feedback_Varyings seems to require a `chars_ptr_array`.
I tried doing it with `chars_ptr` which compiled but the application crashed with a GL error message.

src/gl/interface/gl-objects-buffers.ads
flyx: The PR still contains the subtype.

rogermc2:
I don't know what happened there! I thought I'd fixed fix the subtype thing! 
I'll fix it now.

rogermc2:
I'm having a problem with `Texture_Buffer_Target`
The following worked with `Texture_Buffer_Target` as a subtype but doesn't seem to get inherited now.
~~~
   procedure Load_Texture_Buffer is new
     GL.Objects.Buffers.Load_To_Buffer (GL.Types.Singles.Matrix4_Pointers);
~~~
The error report expects`Buffer_Target` but objects to `Texture_Buffer_Target`.
What do I need to do to get this to work?
An additional generic or additional `procedure  Load_To_Buffer (Target : Texture_Buffer_Target; ...` under the current generic or something else?

A similar problem occurs with
~~~
   procedure Load_Vertex_Buffer is new
     GL.Objects.Buffers.Load_To_Buffer (GL.Types.Singles.Vector2_Pointers)
~~~

flyx:
In the generic subroutines, simply substitute `Target : Buffer_Target` with `Target : Buffer_Target'Class`.

rogermc2:
That fixed it. Thanks.

src/gl/interface/gl-objects-buffers.ads
```diff
@@ -17,17 +18,19 @@ package GL.Objects.Buffers is
...
       with package Pointers is new Interfaces.C.Pointers (<>);
-   procedure Load_To_Buffer (Target : Buffer_Target;
+   procedure Load_To_Buffer (Target : Buffer_Target'Class;
                              Data   : Pointers.Element_Array;
                              Usage  : Buffer_Usage);
 
    -- Use this instead of Load_To_Buffer when you don't want to copy any data
-   procedure Allocate (Target : Buffer_Target; Number_Of_Bytes : Long;
+   procedure Allocate (Target : Buffer_Target'Class; Number_Of_Bytes : Long;
                        Usage  : Buffer_Usage);
```
flyx: This change is unnecessary since Allocate is not a generic procedure and therefore will be inherited for `Texture_Buffer_Target`. Instead, add `'Class` to the generic subroutines below (`Map`, `Pointer`, `Set_Sub_Data`), they need it.

rogermc2: OK
Hopefully correct this time.

flyx:
Great! I will try and do away with those `New_String` allocations from `Interfaces.C.Strings` in the next days. I have unnecessarily used them myself in some places.

flyx:
I pushed a modified `Transform_Feedback_Varyings` and on the way did away with other usages of `New_String` in the project, some of which caused *serious* memory leaks. Please check whether it works for you.

rogermc2:
Using
~~~
      Varyings       : constant String := "position_out,velocity_out";
~~~

In the following, Transform_Feedback_Varyings executes without failure but the subsequent link fails.

~~~
      GL.Objects.Programs.Use_Program  (Update_Program);
      Transform_Feedback_Varyings (Update_Program, Varyings, Interleaved_Attribs);
      Update_Program.Link;
      if not GL.Objects.Programs.Link_Status (Update_Program) then
         Put_Line ("Setup, Update_Program Link failed");
         Put_Line (GL.Objects.Programs.Info_Log (Update_Program));
      else
         Put_Line ("Setup, Update_Program Link ok");
      end if;
~~~
Error messages:
~~~
ERROR: Could not find transform feedback binding for 'position_out.'
ERROR: Could not find transform feedback binding for 'position_out.'
ERROR: Could not find transform feedback binding for 'velocity_out.'

raised GL.ERRORS.INVALID_OPERATION_ERROR : gl-raise_exception_on_opengl_error.adb:10
~~~
I'll try and find if this caused by my code.

rogermc2:
While investigating the `Transform_Feedback_Varyings` procedure I came across a problem with  `Get_Transform_Feedback_Varying` in that some parameters needed to be declared `out` which I have fixed.
However, `Get_Transform_Feedback_Varying` now fails returning an invalid value error.
The functions `Transform_Feedback_Varyings` and `Transform_Feedback_Varying_Max_Length` do not crash but both return the value `0` suggesting that `Transform_Feedback_Varyings` is not working properly.
I have circumvented `Get_Transform_Feedback_Varying` failure by incorporating `Transform_Feedback_Varyings` and `Transform_Feedback_Varying_Max_Length` tests as its possible that a `Transform_Feedback_Varying` having these parameters may be the cause of its failure.
So far, I've not been able to discover other clues as to why linking following `Transform_Feedback_Varyings`  is failing.
Should I submit my proposed changes to  `Get_Transform_Feedback_Varying` in my current open PR or leave it to my next (final) PR?

flyx:
As the current PR does not touch these functions, please put the changes in the final PR.

flyx:
I am unsure whether the points at the end of the varying's names are part of the error message or hint to an error in my code. The fact that `position_out` is mentioned twice hints to an error in my code, I'll investigate (later).

flyx:
I pushed a fix for the wrong number of varyings. Not sure where the dot comes from though.

rogermc2:
I did not receive the usual e-mail advice and can't find any sign of the fix.
However, merging updated master into my test program produces:
~~~
V_Type: SEPARATE_ATTRIBS   V_Length:  2   Max Length:  0
Setup, Update_Program Link failed
ERROR: Could not find transform feedback binding for 'position_out.'
ERROR: Could not find transform feedback binding for 'velocity_out.'
~~~
where,
~~~
Max_Length := Transform_Feedback_Varying_Max_Length (Update_Program);
~~~
So, by `V_Type: SEPARATE_ATTRIBS` and   `V_Length:  2 `, it looks like varyings are being loaded.

flyx:
Indeed. I have no idea where the dots come from, My code only replaces commas with null characters. There is no possibility of a dot getting inserted. Did this work previously for you with the `New_String` code?

flyx:
I noticed that the error says „Could not find transform feedback *binding*“. Perhaps the error is not the varying's name, but the [buffer binding](https://www.khronos.org/opengl/wiki/Transform_Feedback#Buffer_binding)?

rogermc2:
I can't remember if I had it working with the New_String code. Probably not.
 I also wondered about the buffer binding and am not sure what the message means.
I checked everything that I could think of related to binding. 
I'll study your reference to  buffer binding to try and understand just what is supposed to be bound to what.
If really desperate, I'll try using the debugger which is always difficult.

flyx:
Do you have corresponding C code that shows everything that needs to be done?

The documentation I linked mentions `glBindBufferRange`, which is currently not implemented by OpenGLAda. From what I understand, you need to create a `Buffer` object and bind it to `Transform_Feedback_Buffer`. Then you need to call `glBindBufferRange` *for each varying output* with the `index` parameter being whatever you specified in GLSL with `xfb_buffer`. That will tell the program linker where to store the values.

rogermc2:
The example that I am working on is from the red book examples code.
It starts with:
~~~
    update_prog = glCreateProgram();
//  vertex and fragment shader code
    vglAttachShaderSource(update_prog, GL_VERTEX_SHADER, update_vs_source);
    vglAttachShaderSource(update_prog, GL_FRAGMENT_SHADER, white_fs);
    static const char * varyings[] =
    {
        "position_out", "velocity_out"
    };
    glTransformFeedbackVaryings(update_prog, 2, varyings, GL_INTERLEAVED_ATTRIBS);
    glLinkProgram(update_prog);
    glUseProgram(update_prog);
~~~
All array and buffer objects are declared, initialized and bound AFTER this code.
I have checked a couple of other examples and they all start like this.
I've tried making these declarations at the beginning but the result was the same.
I recoded using New_String and got the same results.
During the recoding my first effort missed the second string.
The resulting error message was:
~~~
ERROR: Could not find transform feedback binding for 'position_out.'
ERROR: Could not find transform feedback binding for '.'
~~~
I searched for other instances of `glTransformFeedbackVaryings` and found that problems do occur often involving corruption of the varying text; not just a dot at the end as in our case.

rogermc2:
Attempting debugger use.
~~~
C_Varyings_Ptr : Interfaces.C.Strings.chars_ptr_array (1 .. Num_Strings);
~~~
`C_Varyings_Ptr (1)` is passed to `API.Transform_Feedback_Varyings` as `"position_out["00"]"`

rogermc2:
I now have some OpenGL code (Mesa) which I can look through to see if I can see the problem.

rogermc2:
I have built the C example using Xcode.
It runs without crashing but so far only a black window is displayed

flyx:
Are you getting the error message you showed above or is that just an example?

btw, you cannot step into `API.Transform_Feedback_Varyings` because that is implemented in the OpenGL implementation, which certainly does not have any debugging symbols available.

rogermc2:
So far I haven't seen the linker error message with the C example which is the transform feedback example that my Ada example is based on .
Currently I'm trying to implement `gltransformfeedbackvarying` but slow work at the moment as I'm not very familiar with C++ or with the XCode debugger.
Hopefully, I will be able to obtain some useful information tomorrow.

rogermc2:
While the program does run, I suspect that I may not be getting a proper display because of the following error message:
~~~
transform_feedback[5102:292514] [General] ERROR: Setting <GLFWContentView: 0x1005ad1f0> as the first responder for window <GLFWWindow: 0x10072dc40>, but it is in a different window ((null))! This would eventually crash when the view is freed. The first responder will be set to nil.
~~~
I haven't been able to figure out how to fix it yet so it would help if you can provide advice; but don't worry if you can't.

rogermc2:
No progress with the C version so far as I can't seem to get the program properly initialised including problems with shader versions.

rogermc2:
I have discovered that the version of `GLFW` installed by `brew` no longer works with all programs.
The latest `master` version needs to be downloaded from `https://github.com/glfw/glfw`, built and installed.
Should I update my `Readme` files at this stage to reflect this change or wait until the situation stabilises?

flyx:
Can you give me more details here? Perhaps a link to a relevant GLFW issue or commit? Why does it *no longer* work, did you see an error in an example that was previously working? And do you know if this affects only macOS or all backends?

I would prefer not to tell the users they have to compiler GLFW from scratch unless necessary. If this issue persists until I release the next version, I will put a note in the release notes or changelog that certain features need the newer GLFW code. However, GLFW 3.3 seems to get released soon, so I hope the problem fixes itself.

rogermc2:
This is the error ( `https://github.com/glfw/glfw/issues/909`):
~~~
2016-11-27 19:33:40.726198 GLFW[37260:6137830] [General] ERROR: 
Setting <GLFWContentView: 0x102a26000> as the first responder for window 
<GLFWWindow: 0x10071d510>, but it is in a different window ((null))! 
This would eventually crash when the view is freed. The first responder will be set to nil.
~~~
Currently, I OpenGLAda doesn't seem to be  affected.  It only seems to occur when I'm building C programs.
The `ogl_tutorials/misc05_picking_easy` Ada example works but its C equivalent has the problem.
Hopefully GLFW 3.3 will include the fixes.
My current observations:
1. brew no longer installs 'libglfw.a'. It only installs `dylibs`.
2. the `master` version I referred to above builds 'libglfw3.a' which I need for building my C examples because it solves the problem.
3. the current OpenGLAda examples need  'libglfw.a'.

So,  OpenGLAda seems safe at the moment.

rogermc2:
I now have `glTransformFeedbackVaryings` and `glGetTransformFeedbackVarying` working with the C version.
The resultant shader program links and validates.
There is no `dot` problem which is confirmed by `glGetTransformFeedbackVarying`.

If no vertex array object has been previously bound `Link` succeeds but `Validate` fails.

PR #121, opened 2018-03-17
Picking example

rogermc2:
This is the last of my OpenGL Tutorials examples.
Its supposed to demonstrate "picking" but I'm not sure that its working like its supposed to.


examples/common/src/utilities.ads
```diff
@@ -29,10 +29,6 @@ package Utilities is
      (Maths.Vector5_Pointers);
    procedure Load_Vector6_Buffer is new GL.Objects.Buffers.Load_To_Buffer
      (Maths.Vector6_Pointers);
-   procedure Load_Vertex_Sub_Buffer is new
-     GL.Objects.Buffers.Set_Sub_Data (GL.Types.Singles.Vector3_Pointers);
-   procedure Load_Vertex_Sub_Buffer is new
-     GL.Objects.Buffers.Set_Sub_Data (GL.Types.Singles.Vector4_Pointers);
```
flyx: Seems strange that a new example removes functionality. Is this not needed for another example?

rogermc2: You are correct, I've restored them and added a couple of other print routines that I use for development.

PR #122, opened 2018-03-21
Instancing example

rogermc2:
This demonstrates Instancing.
This example does not use varyings.


examples/common/src/maths.ads
```diff
@@ -43,11 +43,15 @@ package Maths is
...
    function Minimum (I1, I2 : Integer) return Integer;
-   function Minimum (I1, I2 : Single) return Single;
+   function Minimum (F1, F2 : Float) return Float;
+   function Minimum (S1, S2 : Single) return Single;
    function Maximum (I1, I2 : Int) return Int;
+   function Maximum (U1, U2 : UInt) return UInt;
    function Maximum (I1, I2 : Integer) return Integer;
-   function Maximum (I1, I2 : Single) return Single;
+   function Maximum (F1, F2 : Float) return Float;
+   function Maximum (S1, S2 : Single) return Single;
```
flyx: I mostly oversaw or ignored this in previous PRs because it was a minor annoyance. But please, remove all the `Minimum` and `Maximum` functions, you do not need them. In Ada, every scalar type (including `range` types like `Integer` and `digits` types like `Float`) has [attributes `'Max` and `'Min`](http://www.adaic.org/resources/add_content/standards/05rm/html/RM-3-5.html#I1565) that do exactly the same thing as your functions. So please replace the usage of these functions with those attributes. Example: `Int'Max (23, 42)`.

rogermc2: OK.
Seems a bit counterintuitive and non-standard programming but I'll comply with your wishes.

flyx: There's a bit of reasoning behind it, mainly that these functions should be available for *all* (even user-defined) such types, therefore they cannot be normal subprograms defined in the Standard package.

I usually do not use the attributes for `'Pred` and `'Succ` on those types for incrementing / decrementing by one since that is actually more code than doing `+ 1`, but for `'Min` and `'Max`, I think we should use what the language provides us with instead of rolling our own.

rogermc2: Fixed

flyx: How did you fix it? The subprograms are still there.

rogermc2: I'm not sure what happened; they've gone locally.
I must have forgotten to push it!
Should be gone now.

flyx:
Are there any more PRs from you incoming in the near future? I am planning to do a new release soon. I remember you had problems with transform feedback; do you still think there are errors in the OpenGLAda code regarding that?

rogermc2:
There's only the one that involves the transform feedback problem.
I have the C version working and forced an error, using an invalid vary name, that resulted in an error message that DID include the dot at the end of the invalid vary name. So the dot is part of the error message and not part of the problem.
I haven't investigated further as, to get more  transform feedback  experience, I've been working on another  transform feedback tutorial example that has turned out out be much more complicated than I expected as it involves `assimp` and ` imagemagick`. I suspect that you might not want this included as an OpenGLAda example due to its complexity involving additional external C packages?
From our experience so far it does look like there may be a problem relating to the  OpenGLAda code  that involves passing string arrays (?) as both methods that we tried produced the same error message.
Can you suggest further steps or checks that I can use to try and isolate the problem?

PR #131, opened 2019-02-25
Added some dynamic functions and Set_Vertex_Attrib_Pointer2

jokesin:
Added some dynamic functions and Set_Vertex_Attrib_Pointer2
Set_Vertex_Attrib_Pointer2 added in order to comfortable use offsets and other values in bytes,
For example if we use Set_Buffer_Sub_Data, we need data in bytes, so we can use it further.
Some code :

declare

            Block_Size    : Int := Get_Size(Vertex_Data.Vertex_Positions) + Get_Size(Vertex_Data.Vertex_Colors);
            Colour_Offset : Int := Get_Size(Vertex_Data.Vertex_Positions);

begin

            GL.Objects.Buffers.Allocate(Array_Buffer, Long(Block_Size),Static_Draw);
            Utilities.Set_Buffer_Sub_Data(Array_Buffer, 0, Vertex_Data.Vertex_Positions);
            Utilities.Set_Buffer_Sub_Data(Array_Buffer, Colour_Offset, Vertex_Data.Vertex_Colors);
            
            GL.Attributes.Set_Vertex_Attrib_Pointer2(vPosition, 4, Single_Type, 0, 0);
            GL.Attributes.Set_Vertex_Attrib_Pointer2(vColour, 4, Single_Type, 0, Colour_Offset);
end;


rogermc2:
Unfortunately Flyx has been silent for many months, so just when your extensive work might get merged seems unpredictable.

flyx:
@rogermc2 Yes, I don't regularly check this project anymore, sorry. Thing is, I currently don't have an Ada development environment set up on my machines, which makes it difficult to help with error inquiries. However, I can still review PRs since wrapper code usually is easy enough to check without having to compile it.

src/gl/implementation/gl-attributes.adb
```diff
@@ -53,6 +53,102 @@ package body GL.Attributes is
...
+                                         Count          : Component_Count;
+                                         Kind           : Numeric_Type;
+                                         Normalized     : Boolean;
+                                         Stride_In_Bytes, Offset_In_Bytes : Size) is
+   begin
+      API.Vertex_Attrib_Pointer (Index, Count, Kind,
+                                 Low_Level.Bool (Normalized),
+                                 Stride_In_Bytes, Offset_In_Bytes);
+      Raise_Exception_On_OpenGL_Error;
+   end Set_Vertex_Attrib_Pointer2;
```
flyx: I think having `Stride` and `Offset` depending on the data type was a bad idea (there is no reason why `Stride` should be a multiple of the type's size), moreover automatic matching to the underlying function also was a bad idea (there's nothing wrong with giving integer values and specifying that they should be converted to floats). And I think the best thing to do is to fix this properly. The name `Set_Vertex_Attrib_Pointer2` is not helping the user understand what's going on. I suggest the following:

 * Keep the original `Set_Vertex_Attrib_Pointer` to not break existing code, comment its deprecation in the `.ads` (sadly, `pragma Obsolescent` is not standard).
 * Add `Set_Vertex_Float_Attrib_Pointer`, `Set_Vertex_Integer_Attrib_Pointer` and `Set_Vertex_Double_Attrib_Pointer` to match the three underlying API functions, and for each of them, use byte offsets and strides.
 * Update the `.spec` file so that those API functions point to their proper wrappers.

This should give you the same set of possibilities as your current solution.

src/gl/interface/gl-types.ads
```diff
@@ -32,6 +32,7 @@ package GL.Types is
    type Double is new C.double;
 
    -- array types
+   type UShort_Array is array (Size range <>) of aliased UShort;
```
flyx: You need to add `pragma Convention (C, UShort_Array);` below.

src/gl/specs/gl-api.spec
```diff
@@ -233,6 +233,17 @@ spec GL.API is
...
+                                        Count          : UInt;
+                                        Index_Type     : Unsigned_Numeric_Type;
+                                        Element_Offset : UInt;
+                                        Base_Vertex    : Int) with
+	Dynamic => "glDrawElementsBaseVertex", Wrapper => "GL.Objects.Buffers.Draw_Elements_Base_Vertex";
+   procedure Primitive_Restart_Index (Index : UInt) with
+    Dynamic => "glPrimitiveRestartIndex", Wrapper => "GL.Objects.Buffers.Primitive_Restart_Index";
+   procedure Vertex_Attrib_Divisor (Index   : GL.Attributes.Attribute;
+                                    Divisor : UInt) with
+    Dynamic => "glVertexAttribDivisor", Wrapper => "GL.Attributes.Vertex_Attrib_Divisor";
```
flyx: I'd like to have proper indentation for the lines starting with `Dynamic =>` (2 spaces more than starting line).

opengl.gpr
```diff
@@ -70,7 +70,7 @@ library project OpenGL is
    end Builder;
 
    package Compiler is
-      Base_Switches := ("-gnat05", "-gnatwa", "-gnatwl", "-gnatwe",
+      Base_Switches := ("-gnat12", "-gnatwa", "-gnatwl", "-gnatwe",
```
flyx: I assume this is because GNAT Community now only supports the latest Ada standard. However, I am not aware that anyone uses this with a different compiler, so I guess we could as well go with it.

rogermc2:
Great to hear that OpenGLAda is still supported.

jokesin:
@flyx thanks for review!in the coming days I'll correct for the comments. I'll keep you informed.

src/gl/specs/gl-api.spec
jokesin: corrected

src/gl/interface/gl-types.ads
jokesin: corrected

src/gl/implementation/gl-attributes.adb
jokesin: According to spec https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glVertexAttribPointer.xhtml 
and especially that :

> type:
Specifies the data type of each component in the array. The symbolic constants GL_BYTE, GL_UNSIGNED_BYTE, GL_SHORT, GL_UNSIGNED_SHORT, GL_INT, and GL_UNSIGNED_INT are accepted by glVertexAttribPointer and glVertexAttribIPointer. Additionally GL_HALF_FLOAT, GL_FLOAT, GL_DOUBLE, GL_FIXED, GL_INT_2_10_10_10_REV, GL_UNSIGNED_INT_2_10_10_10_REV and GL_UNSIGNED_INT_10F_11F_11F_REV are accepted by glVertexAttribPointer. GL_DOUBLE is also accepted by glVertexAttribLPointer and is the only token accepted by the type parameter for that function.

I decided to do Set_Vertex_Attrib_Pointer with full declaration, keep the original function with warning massage, and add Set_Vertex_Integer_Attrib_Pointer and Set_Vertex_Double_Attrib_Pointer as you advised.
What do you say?

flyx:
Thanks!

Be aware that I squashed your commits, so if you want to continue working on your fork, you should force-update your `develop` from my `develop`. Create separate PR branches to avoid this in the future.

Semi-related: I might try and allocate some time on the weekend to release a new OpenGLAda version since GNAT Community requires the `-gnat12` switch, so your changes may find their way into an release rather quickly. If you have plans to add more stuff, please tell me so I can plan ahead :).

jokesin:
Thanks,Felix!
Ok, I'll take your adviсe.
This time i have no any more stuff to add)

>Вторник,  5 марта 2019, 21:42 +03:00 от flyx <notifications@github.com>:
>
>Thanks!
>Be aware that I squashed your commits, so if you want to continue working on your fork, you should force-update your develop from my develop. Create separate PR branches to avoid this in the future.
>Semi-related: I might try and allocate some time on the weekend to release a new OpenGLAda version since GNAT Community requires the -gnat12 switch, so your changes may find their way into an release rather quickly. If you have plans to add more stuff, please tell me so I can plan ahead :).
>—
>You are receiving this because you authored the thread.
>Reply to this email directly,  view it on GitHub , or  mute the thread .

-------------------------------------
С уважением,
Георгий

rogermc2:
I've just pulled the latest.
gl-text.abd compilation produces a warning that `GL.Attributes.Set_Vertex_Attrib_Pointer` is deprecated.
Should I now update gl-text.abd and any other obsolescent procedures that I find to the newer versions?

flyx:
@rogermc2 I'm taking care of upgrading all the examples.

rogermc2:
Thanks

flyx:
@jokesin I removed your `Draw_Arrays_Instanced` from `GL.Objects.Buffers` because it was already defined in `GL.Objects.Vertex_Arrays`. I also renamed `Primitive_Restart_Index` to `Set_Primitive_Restart_Index` (to conform with naming of similar procedures) and moved it to `GL.Objects.Vertex_Arrays` since it belongs with the other functions.

`glDrawArraysInstanced` does not necessarily require vertex arrays, but my understanding is that this is the *modern* way to use it and therefore it belongs to `GL.Objects.Vertex_Arrays`. I may be wrong here.

PR #138, opened 2019-10-16
Framebuffer

rogermc2:
Additional routines needed to implement Transform Feedback facilities.
Added Transform Buffer example to redbook examples
Added Frustum_Matrix function to common Maths
Also minor modifications to some examples.


rogermc2:
Please delay merging this PR as I am working on another red book example that requires more OpenGL procedures.

rogermc2:
I am suddenly having linking trouble with undefined symbols.
I have attempted to implement `glBindImageTexture`, `glDispatchCompute` and `glMemoryBarrier` but linking fails.
If you have time, could you suggest what might be causing my problem?
As I know you don't have the time to get involved in such problems, I have posted a question in `comp.lang.ada`.

~~~
Undefined symbols for architecture x86_64:
  "_glBindImageTexture", referenced from:
      _gl__objects__textures__bind_image in libOpenGLAda.a(gl-objects-textures.o)
  "_glDispatchCompute", referenced from:
      _gl__objects__programs__dispatch_compute in libOpenGLAda.a(gl-objects-programs.o)
  "_glMemoryBarrier", referenced from:
      _gl__objects__programs__memory_barrier in libOpenGLAda.a(gl-objects-programs.o)
~~~

rogermc2:
I've now completed my current additions.
Please inspect and merge at your convenience.
Thanks,
Roger

src/gl/implementation/gl-objects-buffers.adb
```diff
@@ -29,6 +29,15 @@ package body GL.Objects.Buffers is
...
    Current_Buffers : Buffer_Maps.Map;
 
+   package Transform_Buffer_Maps is new Ada.Containers.Indefinite_Hashed_Maps
+      (Key_Type     => Low_Level.Enums.Buffer_Kind,
+       Element_Type => Transform_Buffer'Class,
+       Hash         => Hash,
+       Equivalent_Keys => Low_Level.Enums."=");
+   use type Transform_Buffer_Maps.Cursor;
+
+   Current_Transform_Buffers : Transform_Buffer_Maps.Map;
```
flyx: I don't see why this would be needed.

src/gl/implementation/gl-objects-buffers.adb
```diff
@@ -46,6 +55,25 @@ package body GL.Objects.Buffers is
...
    end Bind;
 
+   procedure Bind_Transform_Feedback (Object : Transform_Buffer'Class) is
+      Target_Kind : constant GL.Low_Level.Enums.Buffer_Kind :=
+                        GL.Low_Level.Enums.Transform_Feedback;
+      Cursor : constant Transform_Buffer_Maps.Cursor
+        := Current_Transform_Buffers.Find (Target_Kind);
+   begin
+      if Cursor = Transform_Buffer_Maps.No_Element or else
+        Transform_Buffer_Maps.Element (Cursor).Reference.GL_Id /= Object.Reference.GL_Id
```
flyx: Why should the procedure check this? Your code basically checks whether the same ID is not already bound, however if it is, the call to `Bind_Transform_Feedback` would simply do nothing, so there is no reason to guard against it. Using `Indefinite_Hashed_Maps` is a significant overhead that is not justified by what you're doing here.

src/gl/implementation/gl-objects-buffers.adb
rogermc2: `Indefinite_Hashed_Maps` changed to `Indefinite_Ordered_Maps`.
`Indefinite`  seems to be needed to allow `Element_Type => Transform_Buffer'Class`

rogermc2:
I also removed (faulty) `query` related code from this branch.
I must have done some initial `query` coding in the wrong branch! One of my typical blunders! 
Thanks for supporting my efforts. Slow responses are definitely not a problem.

rogermc2:
I think I've addressed your change request.
See my 3 commits on 1 Dec 2019 and following comment also on  1 Dec 2019.
I don't know why these commits and comment appear before the Changes request that they address?

flyx:
My change request was to remove the `Current_Transform_Buffers`, not to change its type. This information is managed by OpenGL, hence we don't need it in the wrapper.

flyx:
What is the package `GL.Enums.Queries` for? I don't see it being used anywhere.

flyx:
I think we have reached a point where I need to think about separating the examples from the main repository. The process should be that we first discuss the changes to the API so that we're both happy with it and *then* the examples can come. Also, the repo seems to be overcrowded for someone who just want to use the library for their own needs. I will see if I can incorporate your changes with the changes I want in the process of splitting into two repos.

flyx:
Merging this for now, I will apply changes afterwards.

flyx:
Okay so I can't compile OpenGLAda anymore since SOIL cannot compile on macOS Catalina, as it depends on the Carbon API which has been removed in Catalina. I am evaluating options, this can take a while.

rogermc2:
Many many thanks for your attention.
I think  the package `GL.Enums.Queries` is used in the remaining branch `query`.
I don't expect to be making any more PRs.
I am now using OpenGLAda extensively on another project which is the project that got me involved with  OpenGLAda in the first place.
I hope that OpenGLAda is being used by other Ada developers.

rogermc2:
I've updated my master but have found that many examples fail due to the buggy 
~~~
 Utilities.Show_Shader_Program_Data (Rendering_Program);
~~~
I thought I'd commented this out of most examples but, apparently, not.
Should I go ahead and comment it out in the examples of my query PR?

flyx:
Please leave the other PR for now; it will be complex to merge as it is.

rogermc2:
OK.
Thanks

PR #139, opened 2019-11-23
Additional support for Transform Feedback

rogermc2:
Adds Query functionality plus an associated example.


flyx:
I review this one the other PR is through since this PR includes the other one and I can't see what it adds that is not in the other one.

flyx:
I integrated your changes to the main repo. The examples are not available in [this separate repo](https://github.com/flyx/OpenGLAda-examples). I added the `feedback_query` example there.

Building the examples has changed in the way that they expect you to install `openglada.gpr` as described in the README. This makes it unnecessary for the user to set all those scenario variables when compiling each example.

I merged the examples README files into the topmost one.

You will notice that I removed the SOIL library completely as it does not compile on macOS anymore. I replaced it with the pure Ada GID library and added a package `GL.Images` which replaces the functionality previously available from the SOIL wrapper. You may need to update production code accordingly.

The examples all compile, using `GL.Images` now if necessary. For all I can see, the content in this PR has been fully integrated and all examples are working on my macOS machine. I dumped the FreeType example since those have been superseded with the official `GL.Text` package as far as I can see.

If you want to make further contributions, I suggest you get a fresh clone of this repo due the examples removal. PRs now need to be made for OpenGLAda and the examples separately.

I will test all this on Windows in the next days and if everything works, craft a new OpenGLAda release.

rogermc2:
Wonderfull news!
Thanks for all your effort.

PR #146, opened 2021-03-11
Add initial support for shader storage buffer objects

Roldak:
This very simple PR enables basic support for shader storage buffer objects ([SSBO](https://www.khronos.org/opengl/wiki/Shader_Storage_Buffer_Object)s).

The `0x90d2` value comes from [here](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_shader_storage_buffer_object.txt).

### DIFF
--- a/src/gl/implementation/gl-low_level-enums.ads
+++ b/src/gl/implementation/gl-low_level-enums.ads
@@ -28,7 +28,8 @@ package GL.Low_Level.Enums is
                         Pixel_Unpack_Buffer, Uniform_Buffer, Texture_Buffer,
                         Transform_Feedback_Buffer, Transform_Feedback,
                         Copy_Read_Buffer, Copy_Write_Buffer,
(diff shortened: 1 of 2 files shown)


flyx:
Code looks good. Is this actually usable without the other things from ARB_shader_storage_buffer_object (I don't follow modern OpenGL these days)? I would like to ensure that all functionality that *is* wrapped can actually be used for something without depending on other functionality that is *not* wrapped.

Roldak:
> Is this actually usable without the other things from ARB_shader_storage_buffer_object (I don't follow modern OpenGL these days)? 

Yes it's usable! Probably some stuff is missing for advanced users but personally I was able to use it as a drop-in replacement for uniform buffer objects where I needed (mainly for storage size reasons).

flyx:
Great, I'll merge this then. Thanks!

PR #147, opened 2021-05-06
Alire manifest

mosteo:
Having a manifest in place enables
- `alr build`ing the library
- `alr publish`ing new versions easily
- Using it as an unversioned dependency in other Alire crates (https://github.com/alire-project/alire/pull/715)

@flyx, I'd understand if you don't want this file in your top-level folder. In that case, it might be still worthy to have it in a separate branch for the above use cases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+alire/
+alire.lock
 lib
 obj
 bin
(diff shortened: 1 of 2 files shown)


flyx:
Since Alire seems to take off, I think this is a good idea. However I would like to restructure the project files so that GID will be an (optional) Alire dependency when doing this. Otherwise, it could lead to problems when people import the GID project which is on Alire separately.

flyx:
This is now obsolete as OpenGLAda has been split into four Alire crates each with a manifest.

PR #149, opened 2022-02-03
Support raw mouse motion

Cre8or:

- Add an API binding to `glfwRawMouseMotionSupported`
- Refactor `Glfw.Enums.Input_Toggle` into two subtypes (cursor and bool toggles)
- Add a window procedure to enable/disable raw mouse motion

Requires Glfw 3.3 or higher.

### DIFF
--- a/opengl-glfw/src/glfw-enums.ads
+++ b/opengl-glfw/src/glfw-enums.ads
@@ -56,9 +56,12 @@ private package Glfw.Enums is
                         Scale_To_Monitor       => 16#2200C#);
    for Window_Info'Size use Interfaces.C.int'Size;
 
-   type Input_Toggle is (Mouse_Cursor);
-   for Input_Toggle use (Mouse_Cursor         => 16#33001#);
+   type Input_Toggle is (Mouse_Cursor, Raw_Mouse_Motion);
+   for Input_Toggle use (Mouse_Cursor         => 16#33001#,
+                         Raw_Mouse_Motion     => 16#33005#);
    for Input_Toggle'Size use Interfaces.C.int'Size;
+   subtype Cursor_Toggle is Input_Toggle range Mouse_Cursor .. Mouse_Cursor;
+   subtype Bool_Toggle is Input_Toggle range Raw_Mouse_Motion .. Raw_Mouse_Motion;
 
    type Joystick_ID is (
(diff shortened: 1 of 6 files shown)


opengl-glfw/src/glfw-enums.ads
```diff
@@ -56,9 +56,12 @@ private package Glfw.Enums is
...
    for Window_Info'Size use Interfaces.C.int'Size;
 
-   type Input_Toggle is (Mouse_Cursor);
-   for Input_Toggle use (Mouse_Cursor         => 16#33001#);
+   type Input_Toggle is (Mouse_Cursor, Raw_Mouse_Motion);
+   for Input_Toggle use (Mouse_Cursor         => 16#33001#,
+                         Raw_Mouse_Motion     => 16#33005#);
    for Input_Toggle'Size use Interfaces.C.int'Size;
+   subtype Cursor_Toggle is Input_Toggle range Mouse_Cursor .. Mouse_Cursor;
+   subtype Bool_Toggle is Input_Toggle range Raw_Mouse_Motion .. Raw_Mouse_Motion;
```
flyx: I don't think keeping `Input_Toggle` as base type makes sense. This type will never be used directly so we could as well have two independent types.

Perhaps the idea was that this way, representation only needs to be declared once, but Ada allows the `'Size` to differ between subtypes, so this is not entirely correct according to the LRM (though GNAT does apply the `'Size` to all subtypes afaik).

Could you change this to two independent types `Cursor_Input_Toggle` and `Bool_Input_Toggle`?

flyx (changes requested):
Thanks for this. If you could address my comment, I'd happily merge this.

opengl-glfw/src/glfw-enums.ads
Cre8or: Indeed, the previous type definition didn't make sense to me either. Perhaps at the time of writing, Glfw didn't support raw mouse motion (yet), so there just wasn't any need for a stricter separation?

Regardless, I implemented your request in [88a7fcf](https://github.com/flyx/OpenGLAda/pull/149/commits/88a7fcfed5efc241a8e14f86bf344328158a62d7). Thoughts?

flyx:
Thanks!

Since this repo is largely dormant between occasional requests, there is no schedule for a new release, which is required if you want to consume this new API through Alire. If that's the case and you don't foresee any additional requests, drop me a message and I can push a new release.

PR #153, opened 2022-07-14
Added Set_Iconification, missing Numpad_Lock key

Cre8or:

- add the window primitive `Set_Iconification`, using the (already implemented) Glfw API procedures `Iconify_Window` and `Restore_Window`
- add a missing value `Numapad_Lock` (key code 282) to `Glfw.Input.Keys.Key`, thus fixing a potential crash when the Numpad lock button is pressed ([gist to reproduce the constraint error)](https://gist.github.com/Cre8or/9e1bfead3726e63644b37e01b8c6c67d).

Requires Glfw 3.3 or higher.

For context, I came across this while binding to ImGui and only discovered the issue through sheer luck (well, through overexcited button mashing resulting in a constraint error). It might be a good idea to take a close look at what keys Glfw defines to see if anything else is missing, though from further button mashing experiments I have been unable to reproduce the issue after my commits. 😛


flyx:
This looks simple enough to merge it directly. Thanks!

Considering the keys, I assume this happened due to GLFW adding support for more keys after I wrote this binding. Sadly, what is „safe“ in C world (adding some new enum value) because you'll just get a value that doesn't match anything you'd expect, is an API break in Ada because it actually checks whether the value is known.

I won't evaluate whether additional keys are missing since am currently considering archiving this project. I can't support it anymore with reasonable effort from an aarch64-darwin system.

